PR #329, opened 2026-04-16
ESD-828: add client-side validation for IX ASN, MAC address, VLAN, and rate limit

Phil-Browne:
IX inputs (ASN, MAC, rate limit, VLAN) had no client-side validation — bad values only failed at the API, after the user had gone through the whole confirmation flow and waited on the call, then got a cryptic error. This validates them up front. Along the way it routes a couple of stragglers through the central validators: VXC rate-limit (was an inline check) and interactive port-name length.

## Changes

- New validators: `ValidateASN` (range 1–4294967295, with `MinASN`/`MaxASN` constants) and `ValidateMACAddress` (EUI-48 via `net.ParseMAC`).
- Wire ASN, MAC, rate limit, and VLAN validation into every IX input path — buy/update flags, JSON, and interactive prompts.
- VXC: replace the inline `rateLimit <= 0` check with `validation.ValidateRateLimit()`.
- Ports: replace the `name == ""` check with `validation.ValidatePortName()` (enforces the 64-char max) in the port and LAG prompts.

## Test plan

- [x] `validation`: new `TestValidateASN`/`TestValidateMACAddress` with boundary cases (min, max, zero, negative, EUI-64 rejection)
- [x] `ix`: new `TestBuildIXRequestFromFlags_ValidationErrors`; updated JSON and prompt tests
- [x] `vxc` and `ports` existing tests pass with the substituted validators
- [x] `golangci-lint run` — 0 issues

### DIFF
--- a/internal/validation/common.go
+++ b/internal/validation/common.go
@@ -41,9 +41,9 @@ const (
 	// ReservedVLAN identifies a VLAN ID that is reserved and cannot be used.
 	ReservedVLAN = 1
 	// MinASN is the minimum valid Autonomous System Number.
-	MinASN = 1
+	MinASN int64 = 1
 	// MaxASN is the maximum valid 32-bit Autonomous System Number.
-	MaxASN = 4294967295
+	MaxASN int64 = 4294967295
 )
 
 // VLANHelpText returns a canonical human-readable description of valid VLAN values,
@@ -202,7 +202,8 @@ func ValidateRateLimit(rateLimit int) error {
 //   - A ValidationError if the ASN is not valid
 //   - nil if the validation passes
 func ValidateASN(asn int) error {
-	if asn < MinASN || asn > MaxASN {
+	v := int64(asn)
+	if v < MinASN || v > MaxASN {
 		return NewValidationError("ASN", asn,
 			fmt.Sprintf("must be between %d and %d", MinASN, MaxASN))
 	}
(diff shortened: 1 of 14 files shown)


internal/validation/common.go
```diff
@@ -185,6 +190,68 @@ func ValidateRateLimit(rateLimit int) error {
...
+//   - nil if the validation passes
+func ValidateASN(asn int) error {
+	v := int64(asn)
+	maxASN := maxSupportedASNForInt()
+	if v < MinASN || v > maxASN {
+		return NewValidationError("ASN", asn,
+			fmt.Sprintf("must be between %d and %d", MinASN, maxASN))
+	}
+	return nil
+}
```
mp1-tomclancy, line 225: `uint32` would be a better way to represent ASNs, it would allow us to remove these magic numbers (which are just the bounds of a `uint32`). It would also allow the entire ASN range to be represented on 32 bit platforms.

mp1-tomclancy (changes requested):
Looks good otherwise, just see my comment on ASNs. You might even be able to remove the validation entirely if `uint32` is used.

internal/validation/common.go
mp1-tomclancy: @Phil-Browne based on our discussion it sounds like this change would need to be co-ordinated across multiple repos, so better to address in the future.

mp1-tomclancy (approved):

PR #330, opened 2026-04-17
ESD-863: split native password prompt into build-tagged file

Phil-Browne:
Moves the `term.ReadPassword`-based `passwordPromptFn` out of `prompts.go` and into a new `prompts_native.go` gated by `//go:build !js || !wasm`, leaving only the uninitialized function pointer in the shared file. This mirrors the existing native/wasm split pattern — `prompts_wasm.go` already wires its implementation via `SetPasswordPrompt` in `init()`, so the native side was the odd one out.

The practical benefit is that `golang.org/x/term` no longer gets pulled into the wasm build, and the platform-specific prompt wiring is consistent across all three prompt functions.

## Notes
- Stacked on top of #329 (depends on the `PasswordPrompt` infrastructure added there). Merge #329 first, then rebase this onto main.
- No behaviour change on native builds — same `term.ReadPassword` path, just relocated.

## Test plan
- [x] `go build ./...`
- [x] `GOOS=js GOARCH=wasm go build -tags js,wasm .`
- [x] `go test ./internal/utils/`
- [x] `golangci-lint run ./internal/utils/`

### DIFF
--- a/internal/utils/prompts_native.go
+++ b/internal/utils/prompts_native.go
@@ -1,0 +1,32 @@
+//go:build !js || !wasm
+// +build !js !wasm
+
+package utils
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/fatih/color"
+	"golang.org/x/term"
+)
+
+func init() {
+	passwordPromptFn = nativePasswordPrompt
+}
+
+func nativePasswordPrompt(msg string, noColor bool) (string, error) {
+	if !noColor {
+		fmt.Print(color.New(color.FgHiRed, color.Bold).Sprint("\U0001f512 " + msg + " "))
+	} else {
+		fmt.Print("\U0001f512 " + msg + " ")
+	}
+
+	password, err := term.ReadPassword(int(os.Stdin.Fd()))
+	fmt.Println() // newline after masked input
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(password)), nil
+}
--- a/internal/utils/prompts_native.go
+++ b/internal/utils/prompts_native.go
@@ -13,7 +13,7 @@ import (
 )
 
 func init() {
-	passwordPromptFn = nativePasswordPrompt
+	SetPasswordPrompt(nativePasswordPrompt)
 }
 
 func nativePasswordPrompt(msg string, noColor bool) (string, error) {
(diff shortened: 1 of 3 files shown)


internal/utils/prompts_native.go
```diff
@@ -0,0 +1,32 @@
+//go:build !js || !wasm
+// +build !js !wasm
```
ianb-mp: This syntax shouldn't be needed since Go 1.17 AFAIK - I think better to leave it off (here and anywhere else it's being used).

internal/utils/prompts_native.go
```diff
@@ -0,0 +1,32 @@
...
+
+func init() {
+	SetPasswordPrompt(nativePasswordPrompt)
+}
+
+func nativePasswordPrompt(msg string, noColor bool) (string, error) {
+	if !noColor {
+		fmt.Print(color.New(color.FgHiRed, color.Bold).Sprint("\U0001f512 " + msg + " "))
+	} else {
+		fmt.Print("\U0001f512 " + msg + " ")
```
ianb-mp, line 23: Add a comment to say what the unicode symbol produces e.g. 
```Go
// lock symbol 🔒
```
(include english description in case output can't render unicode properly)

internal/utils/prompts_native.go
```diff
@@ -0,0 +1,32 @@
+//go:build !js || !wasm
```
ianb-mp: This will be triggered if not js OR not wasm. That doesn't seem right - shouldn't it be not js AND not wasm?

ianb-mp (changes requested):

internal/utils/prompts_native.go
Phil-Browne: Removed the legacy `// +build` line — `//go:build` alone is sufficient on Go 1.17+. Fixed in bad3154.

internal/utils/prompts_native.go
Phil-Browne: Added a comment naming the lock symbol (🔒) so the source remains readable when the glyph cannot render. Fixed in bad3154.

internal/utils/prompts_native.go
Phil-Browne: Good catch — switched to `//go:build !js && !wasm` for clarity and to match the convention used elsewhere in the repo (e.g. `cmd/megaport/megaport.go`, `internal/wasm/stub.go`). Both forms exclude the `GOOS=js GOARCH=wasm` build, but `&&` reads as "neither js nor wasm" which is what we want. Fixed in bad3154.

PR #331, opened 2026-04-17
ESD-829: surface config default-load failures instead of failing silently

Phil-Browne:
## Summary

`applyDefaultSettings` in `cmd/megaport/megaport.go` silently returned on config-manager load failure and early-exited on per-flag `Set` failures via `log.Printf` buried under Go's default logger prefix. Users whose saved defaults (e.g. `output: json`, `no-color: true`) weren't being applied got no signal.

This change routes both failure paths through `output.PrintWarning`:

- Config-load failure emits `Could not load saved default settings: <err>` and returns (defaults stay best-effort — the command still runs).
- Per-flag apply failures are collected and surfaced as a single aggregated warning `Could not apply saved defaults for: <flag-list>` so one bad flag doesn't silently skip the rest.

The five near-identical per-flag blocks are also collapsed into two local closures (`applyBool`, `applyString`).

## Test plan

- [x] `go test -v ./cmd/megaport/` — new `TestApplyDefaultSettings_WarnsOnConfigLoadFailure` plus all existing tests pass.
- [x] Manual smoke: `MEGAPORT_CONFIG_DIR=/dev/null/nope ./megaport version` emits the warning to stdout then runs normally.
- [x] Type-mismatch behavior preserved (config corruption doesn't block the CLI).


cmd/megaport/megaport.go
```diff
@@ -48,6 +47,7 @@ func init() {
 			return fmt.Errorf("invalid output format: %s. Must be one of: %s",
 				outputFormat, strings.Join(utils.ValidFormats, ", "))
 		}
+		output.SetOutputFormat(format)
```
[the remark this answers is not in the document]
ianb-mp: Agreed

cmd/megaport/megaport.go
```diff
@@ -156,82 +163,99 @@ func init() {
...
-				}
-				noPager = boolVal
-			}
+	// --quiet and --verbose are mutually exclusive (see MarkFlagsMutuallyExclusive).
+	// A saved default can combine with a CLI flag (or a manually-edited config
+	// can set both) and bypass cobra's validation. Resolve by preferring the
+	// explicitly-set flag; if both are from config, drop verbose as the safer
+	// default so automation is not unexpectedly chatty.
+	if quiet && verbose {
+		dropped := "verbose"
```
ianb-mp: It might be safer to default to dropping `quiet` - that way if the user is expecting nothing and gets something, they are alerted to a problem whereas the other way around could go unnoticed.

cmd/megaport/megaport_test.go
```diff
@@ -81,6 +83,96 @@ func TestNoPagerDefaultApplied(t *testing.T) {
...
+		output.ResetState()
+		quiet = false
+		verbose = false
+		_ = rootCmd.PersistentFlags().Set("quiet", "false")
+		_ = rootCmd.PersistentFlags().Set("verbose", "false")
+	}()
+
+	warnings := applyDefaultSettings(rootCmd)
+
+	assert.True(t, quiet, "quiet should remain set (safer default)")
```
ianb-mp: See earlier comment about verbose being a safer default.

ianb-mp (changes requested):

cmd/megaport/megaport_test.go
Phil-Browne: Addressed in https://github.com/megaport/megaport-cli/pull/331/commits/ac4d13a0f318611c76452e2aa1460f43c923bf6d

cmd/megaport/megaport.go
Phil-Browne: Addressed in https://github.com/megaport/megaport-cli/pull/331/commits/ac4d13a0f318611c76452e2aa1460f43c923bf6d

ianb-mp (approved):

PR #333, opened 2026-04-19
refactor: add output.SetConfig API and sharpen utils flag-var comment

Phil-Browne:
## Summary

Addresses the "reduce global state in output / utils" cleanup with a minimal, low-risk refactor rather than the full struct-and-single-mutex rewrite originally sketched. The full rewrite was rejected for two concrete reasons:

1. **Cobra flag binding is incompatible with `sync/atomic`.** `PersistentFlags().StringVar(&utils.Env, ...)` etc. require `*string` / `*bool` / `*int`. Switching those globals to atomic types would need per-flag `pflag.Value` shims and would break ~60 direct test assignments in `auth_test.go` / `login_test.go`.
2. **`currentOutputFormat` and `currentVerbosity` are already `atomic.Value`.** Folding them under a shared `sync.RWMutex` would regress spinner-goroutine read performance for no gain.

So instead:

- **`output.SetConfig(OutputConfig)` + `output.DefaultOutputConfig()`**: a single call site for callers that currently invoke several setters in a row. `ResetState()` becomes a one-liner delegating to these. The per-field mutexes stay — `SetConfig` is not atomic across fields, and the godoc says so explicitly.
- **Utils comment**: the existing block comment only justified `Env` and implied the others by reference. Expanded to spell out the per-invocation happens-before invariant across all five variables, acknowledge the WASM re-entry case, and explain why atomics were rejected — so a future author thinking of mutating these at runtime has the full picture.

No dependencies added. No behavioural change.

## Files touched

- `internal/base/output/common.go` — new `OutputConfig` struct, `DefaultOutputConfig()`, `SetConfig()`; `ResetState()` delegates to them.
- `internal/utils/utils.go` — comment-only: expanded the happens-before rationale covering all five flag-bound globals.
- `internal/base/output/output_test.go` — `TestDefaultOutputConfig_Values`, `TestSetConfig_AppliesAllFields`, `TestSetConfig_ZeroValueClearsState`, `TestResetState_ViaDefaultOutputConfig`.
- `internal/base/output/pager_test.go` — `TestSetConfig_NoPager` (native-only, since pager state is native-only; WASM provides stubs).

## Test plan

- [x] `go test ./...` passes
- [x] `golangci-lint run` reports 0 issues
- [x] Code review run — three comments applied (drop empty-string default substitution from SetConfig, rename `TemplateStr` → `Template`, broaden utils future-change guidance)
- [x] Security audit run — Low risk, no findings (no new I/O / dependencies / credential handling / file paths / network calls)
- [x] Copilot review feedback addressed: added direct test coverage for the new API, corrected WASM pager-stub commentary, updated this description to list the test files.

### DIFF
--- a/internal/base/output/common.go
+++ b/internal/base/output/common.go
@@ -27,10 +27,17 @@ var (
 // in fields will be included in output. Pass nil to restore full output (all fields).
 // This function is goroutine-safe. Tests should call defer SetOutputFields(nil) to
 // reset state between test cases.
+//
+// The slice is copied before being stored, so callers may mutate or reuse
+// the backing array after the call returns without affecting the filter.
 func SetOutputFields(fields []string) {
 	outputFieldsMu.Lock()
 	defer outputFieldsMu.Unlock()
-	outputFields = fields
+	if fields == nil {
+		outputFields = nil
+		return
+	}
+	outputFields = append([]string(nil), fields...)
 }
 
 // getOutputFields returns a copy of the current field filter under a read lock.
@@ -136,8 +143,10 @@ func getNoHeader() bool {
 // sentinel. Callers that want defaults should start from
 // DefaultOutputConfig() and override the fields they care about.
 type OutputConfig struct {
-	// Fields is stored by reference by SetOutputFields; callers must not
-	// mutate the backing array after passing it through SetConfig.
+	// Fields is the list of field names (json tag or header display name,
+	// case-insensitive) to include in output. Nil or empty means all fields.
+	// The slice is copied by SetOutputFields, so callers may reuse the
+	// backing array after SetConfig returns.
 	Fields   []string
 	Query    string
 	NoHeader bool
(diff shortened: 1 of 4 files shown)


Phil-Browne:
Good catch @copilot — fixed in e059771. `GetNoPager` is defined on WASM as a stub (always returns false); the real reason the round-trip test is native-only is that pager *state* is native-only. Updated both comments to say so.

Phil-Browne:
Fair point @copilot — PR body updated. Now lists all four files with a one-line description each, plus a note that the test additions came in from the review cycle.

Phil-Browne:
Both fixed in a3456fa:

- `SetOutputFields` now copies the input slice before storing it. Reads already returned a copy; writes didn't. With both sides copying, the `OutputConfig.Fields` godoc no longer needs the "do not mutate after passing" warning — the aliasing risk is gone regardless of which API (individual setter or `SetConfig`) the caller uses.
- `SetConfig` godoc now correctly names the mechanism per field: mutexes for Fields/Query/NoHeader/NoPager/Template, `atomic.Value` for Format/Verbosity. The non-atomic-across-fields claim follows from an accurate premise.

Phil-Browne:
Both addressed in eb60e9c:

- Unified the nil/empty wording on `SetOutputFields` to match `OutputConfig.Fields` ("nil or an empty slice restores full output"). Behaviour was already identical — callers gate on `len(fields) > 0` — the comment just lagged.
- Added two regression tests: `TestSetOutputFields_CopiesInputSlice` and `TestSetConfig_CopiesFieldsSlice`. Both mutate the caller's slice after the call and assert the stored filter is unchanged, so the aliasing hole can't silently come back.

ianb-mp (approved):

PR #341, opened 2026-04-25
ESD-893: Fix output formatting inconsistencies

Phil-Browne:
Replace `fmt.Printf` calls in `status_output.go` and `mcr_prompts.go` with `output.PrintPlain` (a new icon-free output helper) so that `--no-color` and pager/WASM output capture are respected consistently. `output.PrintInfo` was considered but adds an "ℹ" prefix unsuitable for section headings and prompt context lines. `PrintPlain` follows the same quiet/JSON-routing rules as `PrintInfo` but writes no icon prefix. Change the terminal width fallback in `table.go` from 100 to the POSIX standard 80, and add `SetTerminalWidthForTesting` to let tests pin the width (with a matching no-op stub in the WASM build).

Also fixes mock interface drift: the dependency on megaportgo has been bumped from **v1.9.0 → v1.10.1** to pick up the `WaitForMCRReady` method added to `MCRService` and 17 new NAT Gateway methods (`CreateNATGatewayPacketFilter`, the packet-filter and prefix-list family, route/BGP helpers). Several mock types were updated to satisfy the extended interfaces; stubs follow the existing `fmt.Errorf("mock: X not configured")` convention.

This PR also adds `output.PrintNewline()` (native + WASM variants) so that the blank-line separators between dashboard sections in `status_output.go` are routed through the output package's capture path rather than going directly to `os.Stdout`, keeping the WASM build consistent.

## Notable changes unrelated to the core fix

**megaportgo v1.10.1 bump** (`go.mod`, `go.sum`, `vendor/`): Required to pick up `WaitForMCRReady` on `MCRService` and the full NAT Gateway surface. The v1.9.0 SDK did not export these types, causing CI typecheck failures.

**Mock completions** (`apply`, `mcr`, `nat_gateway`, `status`, `topology`): Stub implementations added for all interface methods added in megaportgo v1.9.0–v1.10.1.

**`SetTerminalWidthForTesting` helper** (`internal/base/output/table.go`): The 80-char fallback exposed eight table output tests that were implicitly relying on the old 100-char width. A new exported helper pins the terminal width during tests; the affected tests are updated to use it. A WASM no-op stub is included in `table_wasm.go`.

**`output.PrintNewline()`** (`internal/base/output/messages_native.go`, `messages_wasm.go`): New helper that wraps `fmt.Println()` on native and writes `"\n"` to `WasmOutputBuffer` on WASM, so blank-line separators are captured in browser builds.

**`output.PrintPlain()`** (`internal/base/output/messages_native.go`, `messages_wasm.go`): New helper that prints a formatted line with no icon/color prefix. Suppressed in quiet mode; routed to stderr in JSON output mode (consistent with `PrintInfo`). Used for dashboard section headings and interactive prompt context lines where an "ℹ" prefix would be misleading or distracting.


ianb-mp (approved):
lgtm

PR #343, opened 2026-04-25
ESD-895: Miscellaneous cleanup

Phil-Browne:
Fixes several minor quality issues across the CLI codebase.

**Deprecated API removed**: `internal/utils/retry.go` was calling `netErr.Temporary()`, which is deprecated and was suppressed with `//nolint:staticcheck`. Replaced with `netErr.Timeout()` only; the existing syscall and substring fallbacks already cover the transient-error cases that `Temporary()` was meant to catch.

**Makefile clean target**: The `clean` target only removed `coverage.out` by exact name, missing `cover_*.out` and `coverage_*.out` files generated by per-package test runs. Updated to `cover*.out coverage*.out`.

**`update-tags` confirmation**: All four `update-tags` commands (ports, VXC, MCR, MVE) replaced existing tags silently. Added a `--force` flag; without it, the command now prompts "This will replace N existing tag(s). Continue? [y/N]" when there are existing tags. Interactive mode is unaffected. Tests updated accordingly, including injected confirm/cancel coverage via the existing `SetConfirmPrompt` hook.

**Mock interface fixes**: `WaitForMCRReady` was added to the `megaportgo` `MCRService` interface in a recent SDK bump but was missing from the topology and MCR test mocks, causing a typecheck failure under `golangci-lint`. Both mocks now implement the method; the MCR mock follows the established pattern with injectable error and captured-ID fields.

**File-read security hardening**: `ParseResourceTagsInput` and `parseResourceTagsInputExtended` passed `--json-file` / `--tags-file` values directly to `os.ReadFile` with no sanitisation. Introduced a `readTagsFile` helper that cleans the path, rejects upward traversal (`../../`), and enforces a 1 MiB size limit before reading. Tests added for both rejection cases.


ianb-mp (approved):
lgtm

PR #345, opened 2026-04-27
docs: Add jCodemunch MCP code exploration policy to CLAUDE.md

Phil-Browne:
Adds a jCodemunch-MCP code exploration policy section to `CLAUDE.md` for the megaport-cli project. This instructs Claude Code to use jCodemunch-MCP tools for all code navigation instead of falling back to `Read`, `Grep`, `Glob`, or `Bash` shell commands.

The added sections cover:

- **Code Exploration Policy** — when to use `search_symbols`, `search_text`, `get_symbol_source`, `get_context_bundle`, `get_file_outline`, and relationship queries (`find_importers`, `get_blast_radius`, `get_changed_symbols`, etc.)
- **Session-Aware Routing** — opening move with `plan_turn`, how to interpret confidence levels, and how to handle negative evidence from searches
- **Model-Driven Tool Tiering** — passing `model=` to `plan_turn` so the server narrows the exposed tool list to match the active model's capabilities

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,3 +91,78 @@ All commands that display data should use the `output` package. Support `--outpu
 - `github.com/spf13/cobra` — CLI framework
 - `github.com/jedib0t/go-pretty/v6` — Table rendering
 - `github.com/stretchr/testify` — Test assertions
+
+
+## Code Exploration Policy
+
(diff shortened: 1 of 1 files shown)


mp1-tomclancy:
I think this might cause problems for contributors who aren't using `jcodemunch-mcp`. I also noticed that this MCP requires a license for commercial use: https://github.com/jgravelle/jcodemunch-mcp/blob/main/LICENSE

PR #351, opened 2026-04-29
ci: harden GitHub Actions workflows (permissions, SHA pins, Dependabot)

Phil-Browne:
## Summary

Hardens the three CI workflows (`test.yml`, `integration-test.yml`, `release.yaml`) against the GitHub Actions script-injection / token-theft class actively exploited in 2025 (HackerBot Claw; same class discussed alongside CVE-2026-3854).

- **Add top-level `permissions: contents: read`** to `test.yml` and `integration-test.yml`. `test.yml` runs on `pull_request` from forks — the default `GITHUB_TOKEN` was over-scoped (write to most resources). Least-privilege is the right default. `release.yaml` already had its permissions explicitly scoped.
- **SHA-pin all third-party actions** (with version comment) instead of mutable major-version tags. Mutable `@v3` / `@v5` / `@v6` references can be silently re-pointed by a compromised maintainer — a high-value target on the **release workflow**, which holds `GPG_PRIVATE_KEY`, `PASSPHRASE`, and `HOMEBREW_TOKEN`.
  - `test.yml`: `actions/checkout` v4.3.1, `actions/setup-go` v5.6.0, `codecov/codecov-action` v4.6.0
  - `integration-test.yml`: `actions/checkout` v4.3.1, `actions/setup-go` v5.6.0
  - `release.yaml`: `actions/checkout` v4.3.1, `actions/setup-go` v5.6.0, `docker/setup-qemu-action` v3.7.0, `docker/setup-buildx-action` v3.12.0, `docker/login-action` v3.7.0, `crazy-max/ghaction-import-gpg` v6.3.0, `goreleaser/goreleaser-action` v5.1.0, `actions/attest-build-provenance` v2.4.0
- **Add `github-actions` ecosystem to Dependabot** so SHA bumps land as reviewable PRs (weekly, grouped, minor/patch only — major bumps stay human decisions).

## What this does *not* change

- No workflow logic, no scope changes to `release.yaml` permissions, no schedule changes to `integration-test.yml`.
- Same Go version, same goreleaser invocation, same Docker steps.

## Threat-model context

The repo's `pull_request` workflow does not use `pull_request_target` and does not unwrap PR-controlled strings (title/body/branch) into `run:` blocks, so there were no live injection findings. This PR closes the surrounding hardening gaps:

| Gap | Before | After |
|---|---|---|
| Token scope on PR runs | implicit (write-most) | `contents: read` |
| Action references (test) | `@v4`/`@v5` | `@<sha>  # v...` |
[...]

PR #356, opened 2026-05-20
ESD-999: Add golangci-lint workflow for pull requests

Phil-Browne:
## Summary

Adds a dedicated GitHub Actions workflow (`.github/workflows/lint.yml`) that runs `golangci-lint` on every pull request and every push to `main`. Previously lint only ran on developer machines, so style/correctness regressions could land on `main` undetected. This closes that CI gap and brings `megaport-cli` in line with the lint coverage that already exists in `megaportgo` and `terraform-provider-megaport`.

## Conventions followed

Picks up the hardening pattern established in #351:

- **`permissions: contents: read`** at the workflow level — `pull_request` runs from forks no longer get a write-scoped `GITHUB_TOKEN`.
- **SHA-pinned third-party actions** with version comments — mutable `@v4` / `@v5` / `@v9` tags can be silently re-pointed.
  - `actions/checkout` v4.3.1, `actions/setup-go` v5.6.0, `golangci/golangci-lint-action` v9.2.0
- **Pinned golangci-lint version** (`v2.12.2`) so CI matches the `version: "2"` schema in `.golangci.yml` and does not drift if a new major lands upstream. Future bumps will arrive as reviewable Dependabot PRs via the `github-actions` ecosystem already configured in `.github/dependabot.yml`.
- `cache: false` on `setup-go` because `golangci-lint-action` manages its own cache — leaving both on causes warnings and occasional cache poisoning between jobs.

## Test plan

- [ ] The lint job on this PR completes successfully (validates the workflow itself)
- [ ] On a follow-up PR that intentionally introduces a lint violation, the lint job fails and blocks merge
- [ ] Dependabot opens a PR for `golangci/golangci-lint-action` when the next minor/patch release lands


.github/workflows/lint.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff  # v5.6.0
+        with:
+          go-version-file: 'go.mod'
+          cache: false
+      - name: Run golangci-lint
+        # v9.x SHAs are not yet on the org Actions allowlist; v8.0.0 is.
+        uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9  # v8.0.0
```
mp1-tomclancy: Update this to:

```
uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20" # v9.2.0
```

(SHA has been added to allowlist)

mp1-tomclancy (changes requested):

.github/workflows/lint.yml
Phil-Browne: Addressed in https://github.com/megaport/megaport-cli/pull/356/commits/6c663a3653602f48fe25470beb669c447c8c6660

mp1-tomclancy (approved):

PR #358, opened 2026-05-20
ESD-1170(feat-mve): support adminPassword for Cisco and Palo Alto vendor configs

Phil-Browne:
## Summary

Surfaces the new `AdminPassword` field that megaportgo PR [#153](https://github.com/megaport/megaportgo/pull/153) added to `CiscoConfig` and `PaloAltoConfig` (shipped in v1.10.3). Without this field the CLI cannot provision Cisco FTDv or Palo Alto MVEs — the staging API rejects the buy order with `400 Validation failed adminPassword is required`.

ESD-1170 is the CLI counterpart to the SDK change.

### Core support
- **JSON parsing** (`mve_inputs.go`)
  - `parseCiscoConfig` reads optional `adminPassword` and passes it to `CiscoConfig.AdminPassword`.
  - `parsePaloAltoConfig` accepts `adminPassword` as an alternative to `adminPasswordHash` and now requires at least one of the two (both `omitempty` in the SDK; the API needs one).
- **Validation** (`internal/validation/mve.go`)
  - `ValidatePaloAltoConfig` mirrors the parser: at least one of `AdminPassword` or `AdminPasswordHash` must be set. Previously forced `AdminPasswordHash` to be non-empty, which would have blocked the new plaintext path.
- **Interactive prompts** (`mve_prompts.go`)
  - Cisco: new optional `Enter admin password` prompt after the FMC NAT ID step.
  - Palo Alto: replaces the always-required hash prompt with two prompts (plaintext password and hash, either is fine, at least one required).
- **Help text** (`mve.go`) — Cisco JSON buy example now shows `adminPassword`.

### Security: no-echo password input
The interactive prompts that collect credentials previously echoed input to the terminal (the existing `adminPasswordHash` prompt did this too, but plaintext passwords are materially worse). Added:

- **`utils.SecretResourcePrompt`** — uses `term.ReadPassword` on a TTY so input is not displayed; falls back to the standard echoed prompt for piped input (CI, scripts). Same `Get`/`Set`/`sync.RWMutex` pattern as the existing prompt vars, so tests override it the same way.
- All three MVE admin-credential prompts (new Cisco `adminPassword`, new Palo Alto `adminPassword`, existing Palo Alto `adminPasswordHash`) now route through `SecretResourcePrompt`.

### Tests
- `TestParseVendorConfig_AdminPasswordPropagation` — round-trips the field through both vendors, and confirms Cisco's `AdminPassword` is empty when the JSON key is absent.
- New `TestParseVendorConfig` rows for cisco-with-password, palo-alto-plaintext-only, palo-alto-hash-only (assert empty plaintext), and palo-alto-missing-both-credentials.
- Updated Palo Alto validation test: removed the "missing hash" expectation, added "missing both" (error) and "plaintext-only" (valid) cases.
- `TestBuyMVE/interactive_mode_success` now exercises the cisco admin-password prompt; the test harness installs the secret-prompt override alongside the resource-prompt override (single shared closure preserves the existing prompt-sequence model).

## Behavioural change to call out

Previously `adminPasswordHash` was **required** for Palo Alto MVEs in JSON, flag-driven, and interactive flows. It is now **optional, with the constraint that at least one of `adminPassword`/`adminPasswordHash` must be supplied**. Existing scripts that provide `adminPasswordHash` keep working without modification.

## Test plan

- [x] `go build -v ./...` — clean
- [x] `go test ./...` — all packages pass
- [x] `golangci-lint run ./...` — 0 issues
- [x] `/code-review` — only Medium findings (test-coverage gaps) which are now addressed in this PR
- [x] `/security-audit` — Medium finding (plaintext echo) addressed via `SecretResourcePrompt`; remaining items are Informational
- [ ] Manual interactive buy of a Cisco FTDv MVE (requires API credentials — not run locally)


internal/commands/mve/mve_prompts.go
```diff
@@ -237,10 +242,17 @@ func promptMVEVendorConfig(vendorStr string, imageID int, productSize string, mv
 		if err != nil {
 			return nil, err
 		}
-		adminPasswordHash, err := utils.ResourcePrompt("mve", "Enter admin password hash (required): ", noColor)
+		adminPassword, err := utils.SecretResourcePrompt("mve", "Enter admin password (optional, leave blank if providing hash): ", noColor)
 		if err != nil {
 			return nil, err
 		}
+		adminPasswordHash, err := utils.SecretResourcePrompt("mve", "Enter admin password hash (optional, leave blank if providing plaintext password): ", noColor)
```
ianb-mp: What happens if both `adminPassword` and `adminPasswordHash` are set, and do we need to check for that?

internal/utils/prompts_secret_native.go
```diff
@@ -0,0 +1,48 @@
+//go:build !js || !wasm
+// +build !js !wasm
```
ianb-mp: Remove the old style build tag

internal/validation/mve.go
```diff
@@ -462,8 +463,8 @@ func ValidatePaloAltoConfig(config *megaport.PaloAltoConfig) error {
 	if config.SSHPublicKey == "" {
 		return NewValidationError("SSH public key", config.SSHPublicKey, "cannot be empty")
 	}
-	if config.AdminPasswordHash == "" {
-		return NewValidationError("admin password hash", config.AdminPasswordHash, "cannot be empty")
+	if config.AdminPassword == "" && config.AdminPasswordHash == "" {
```
ianb-mp, line 466: I noticed this is checked in `mve_inputs.go` and also in `mve_prompts.go` - is it necessary to check in all three places?

ianb-mp (changes requested):
A couple of minor issues

Phil-Browne:
Thanks @ianb-mp — all three points are addressed in 10fb5aa.

**1. What happens if both `adminPassword` and `adminPasswordHash` are set?**

The API spec is "exactly one" — accepting both leaves which credential the platform actually applies undefined, so I treated it as a validation failure rather than a silent pick:

- `ValidatePaloAltoConfig` now rejects the "both set" case alongside the existing "neither set" case.
- Interactive prompts (`mve_prompts.go`) skip the hash prompt entirely when the user has typed a plaintext password, so the interactive flow can't even produce a "both set" config.
- JSON parsing still accepts both fields verbatim and lets validation surface the error — keeps `parsePaloAltoConfig` purely structural.

**2. `internal/utils/prompts_secret_native.go` — remove the old-style build tag**

Done — dropped the `// +build !js !wasm` line. `//go:build !js || !wasm` is on its own now, matching the convention in the rest of the repo.

**3. `mve_inputs.go`, `mve_prompts.go`, and `validation/mve.go` all checking the same rule**

Agreed it didn't need to live in three places. `ValidateMVEVendorConfig` runs against the built request in both `BuyMVE` and `ValidateMVE` paths, so it covers every entry mode (interactive, flags, JSON). I removed the duplicate guards from `mve_inputs.go` and `mve_prompts.go` and left `ValidatePaloAltoConfig` as the single source of truth. The existing parse / prompt tests that asserted the early error were updated (parsing now accepts neither/either/both, the interactive flow leans on validation), and a new "both set" case was added to `validation/mve_test.go`.

`go test ./...` and `golangci-lint run ./...` are both clean.

ianb-mp (approved):

PR #363, opened 2026-05-21
ESD-1184: Default all cancel commands to immediate, opt-in deferred via --later

Phil-Browne:
Cancellation behavior was inconsistent: `ports delete` cancelled immediately, but `vxc`/`ix delete` defaulted to deferred (end-of-billing-cycle), and `mcr`/`mve delete` carried a `--now` flag the SDK ignored. This makes every delete command immediate by default, with deferred as an explicit opt-in only where the API actually supports it.

| Command | API supports deferred? | Default | Opt-in flag |
|---|---|---|---|
| `ports delete` | No | now | — |
| `mcr delete` | No | now | — |
| `mve delete` | No | now | — |
| `vxc delete` | Yes | now | `--later` |
| `ix delete` | Yes | now | `--later` |

- Original ESD-1184 scope: the API/SDK no longer accept `DeleteNow=false` for ports, so `DeletePort` hardcodes immediate and the `--now` flag is gone.
- `mcr`/`mve delete` drop their no-op `--now` flag and hardcode immediate too.
- `vxc`/`ix delete` now default to immediate and take `--later` for end-of-cycle cancellation.
- New cmdbuilder helpers `WithImmediateDeleteFlags`/`WithDeferredDeleteFlags` declare each command's intent at the flagset level.
- Bumps `megaport/megaportgo` v1.10.3 → v1.11.0 (ports cancellation change + `DeletePort` nil-guard).

## Breaking changes

- `--now` is gone from `ports`/`mcr`/`mve delete` and now errors with `unknown flag: --now`. Behavior is unchanged — those commands were already immediate.
- `vxc delete` / `ix delete` now cancel **immediately** by default instead of at end-of-cycle. Scripts relying on the old default must add `--later`.

## Test plan

- [x] Build, vet, `go test ./...`, `golangci-lint` all clean
- [x] `--help` confirms `--now` gone from ports/mcr/mve and `--later` present on vxc/ix
- [x] Unit tests assert `DeleteNow=true` by default on VXC/IX and that `--later` flips it
- [ ] Staging: each `delete --force` sends CANCEL_NOW; `vxc delete --later --force` schedules end-of-term

### DIFF
--- a/docs/guides/aws-direct-connect.md
+++ b/docs/guides/aws-direct-connect.md
@@ -156,9 +156,9 @@ To remove the connection when no longer needed:
 
 ```sh
 # Delete VXC first
-megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
+megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 
-# Then delete the port (ports are always deleted immediately)
+# Then delete the port
 megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 ```
 
--- a/docs/guides/aws-direct-connect.md
+++ b/docs/guides/aws-direct-connect.md
@@ -158,8 +158,8 @@ To remove the connection when no longer needed:
 # Delete VXC first
 megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
 
-# Then delete the port
-megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
+# Then delete the port (ports are always deleted immediately)
+megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 ```
 
 ## Related Commands
--- a/docs/guides/portal-to-cli-migration.md
+++ b/docs/guides/portal-to-cli-migration.md
@@ -170,13 +170,13 @@ VXCs must be deleted before deleting their parent port or MCR:
 
 ```sh
 # Delete VXCs first
-megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
+megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 
-# Then delete the port (ports are always deleted immediately)
+# Then delete the port
 megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 ```
 
-> **Note:** For VXCs and other deferred-cancellation resources, omit `--now` to schedule deletion at the end of the current billing period. Ports no longer support deferred cancellation and are always deleted immediately.
+> **Note:** All `delete` commands now cancel resources immediately by default — billing stops right away. VXCs and IXs are the only resource types that still accept deferred cancellation; pass `--later` to schedule cancellation for the end of the current billing cycle instead.
 
 ### Manage billing contact
 
@@ -319,7 +319,7 @@ VXCS=$(megaport-cli vxc list --a-end-uid "$PORT_UID" --output json \
 
 for uid in $VXCS; do
   echo "Deleting VXC ${uid}..."
-  megaport-cli vxc delete "$uid" --now --force --quiet
+  megaport-cli vxc delete "$uid" --force --quiet
 done
 
 echo "Deleting port ${PORT_UID}..."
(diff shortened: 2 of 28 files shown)


internal/commands/ports/ports.go
```diff
@@ -245,16 +245,17 @@ func buildPortManagementCommands(rootCmd *cobra.Command) (list, get, status, del
...
 		WithArgs(cobra.ExactArgs(1)).
 		WithColorAwareRunFunc(DeletePort).
-		WithSafeDeleteFlags().
-		WithLongDesc("Delete a port from your account in the Megaport API.\n\nThis command allows you to delete an existing port by providing the UID of the port as an argument. By default, the port will be scheduled for deletion at the end of the current billing period.").
-		WithOptionalFlag("now", "Delete the port immediately instead of waiting until the end of the billing period").
+		WithBoolFlagP("force", "f", false, "Skip confirmation prompt").
+		WithBoolFlag("safe-delete", false, "Fail if the resource has attached VXCs or other active services").
+		WithLongDesc("Delete a port from your account in the Megaport API.\n\nThis command deletes an existing port by providing the UID of the port as an argument. Ports are deleted immediately; deferred cancellation at the end of the billing period is no longer supported.").
 		WithOptionalFlag("force", "Skip the confirmation prompt and proceed with deletion").
+		WithOptionalFlag("safe-delete", "Fail if the resource has attached VXCs or other active services").
```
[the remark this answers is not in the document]
Phil-Browne: Fixed in a8fda52. Added `WithImmediateSafeDeleteFlags()` to `internal/base/cmdbuilder/flagsets.go` (mirrors the existing `WithSafeDeleteFlags` pattern but omits `--now`) and the ports delete command now uses it instead of inlining `--force`/`--safe-delete`.

docs/guides/portal-to-cli-migration.md
```diff
@@ -172,11 +172,11 @@ VXCs must be deleted before deleting their parent port or MCR:
...
 megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
 
-# Then delete the port
-megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
+# Then delete the port (ports are always deleted immediately)
+megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 ```
 
-> **Note:** Omit `--now` to schedule deletion at the end of the current billing period instead of immediately.
+> **Note:** For VXCs and other deferred-cancellation resources, omit `--now` to schedule deletion at the end of the current billing period. Ports no longer support deferred cancellation and are always deleted immediately.
```
mp1-tomclancy: This could lead to confusion, due to a inconsistency in how the flags are applied between resources. e.g.

`ports delete` + no flag -> **now**
`vxc delete` + no flag -> **later**

If deferred delete is the special case, then we should default to **now** and provide an optional `--deferred` or `--later`

docs/guides/aws-direct-connect.md
```diff
@@ -158,8 +158,8 @@ To remove the connection when no longer needed:
 # Delete VXC first
 megaport-cli vxc delete vxc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
 
-# Then delete the port
-megaport-cli ports delete port-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx --now
```
mp1-tomclancy, line 162: Unrelated, but I just noticed that `ports` is plural whereas other resources like `vxc` are singular.

Phil-Browne: Intentional to maintain scope, but considering changing in the future

docs/guides/portal-to-cli-migration.md
Phil-Browne: Good catch — fixed in 02fbf07.

`vxc delete` and `ix delete` now default to immediate cancellation and accept an opt-in `--later` flag to schedule end-of-billing-cycle cancellation. That matches what `ports delete` does after this PR, so all delete commands behave the same when invoked with no flag.

While I was in there, `mcr delete` and `mve delete` also had a `--now` flag that did nothing useful — the SDK rejects `DeleteNow=false` for both. Dropped the flag from both commands and hardcoded immediate deletion.

Net result of the CLI delete-flag rules across all product types:

| Command | API supports deferred? | Default | Opt-in flag |
|---|---|---|---|
| `ports delete` | No | now | — |
| `mcr delete` | No | now | — |
| `mve delete` | No | now | — |
| `vxc delete` | Yes | now | `--later` |
| `ix delete` | Yes | now | `--later` |

Examples, generated docs, README, and migration guide updated to match. Tests assert the new defaults and the `--later` behavior on VXC/IX.

Phil-Browne: Addressed in https://github.com/megaport/megaport-cli/pull/363/commits/02fbf07e36d33f8c818bdeb630796e6ab84a9800

mp1-tomclancy (approved):

PR #364, opened 2026-05-21
test(ports): migrate integration tests to direct-action pattern

Phil-Browne:
Reintroduces a ports integration suite (ESD-1174), replacing the `exec.Command`-based suite under `internal/test/` that #361 removed. The new tests live in their proper place — `internal/commands/ports/ports_integration_test.go`, build tag `integration`, same package — and call the action functions directly (Approach B).

Four lifecycle scenarios, each `t.Parallel()`: single 1G port, 10G LAG, `--json` round-trip, and `--json-file` round-trip (buy → get → list → update → get). Real parallelism brings the wall time to ~139s, down from ~337s on the old serialized baseline.

## Making it parallel-safe

Two pieces of global state made a naive `t.Parallel()` unsafe; both are worth understanding for review:

- **Login.** New `testutil.RequireSharedIntegrationClient` authorizes once per process (`sync.Once`) and installs `config.SetLoginFunc` exactly once. Per-test save/restore pairs would race on the global `config.LoginFunc` and leave it pointing at a stale closure. `LoginWithClient` stays for serial read-only suites (e.g. `locations`).
- **Output capture.** Side-effecting actions (`BuyPort`, `UpdatePort`, …) run without `output.CaptureOutput`, which swaps the process-wide `os.Stdout` under a mutex — under `t.Parallel()` that both serializes every wrapped call and races stdout against other tests' spinner goroutines. The `BuyPort` SDK response is instead recovered via an `init()` hook on `buyPortFunc` into a `sync.Map`. State assertions read through `testutil.SharedIntegrationClient(t)` (direct SDK calls) rather than scraping `GetPort`/`ListPorts` stdout.

## Cleanup

Registered via `t.Cleanup`, so deletion runs even on failure. After `DeletePort` it polls the SDK directly (2s backoff, 60s deadline) until the port reports `DECOMMISSIONING`/`DECOMMISSIONED` — the API takes a few seconds to reflect the transition and exposes no completion signal. This poll loop holds the only `time.Sleep` in the suite; the buy/update paths block on the SDK's `WaitForProvision`/`WaitForUpdate`.

## Notes

- Resource names are prefixed `CLI-Test-` (per `docs/INTEGRATION_TESTING.md`) so cleanup tooling can find leaks; staging location `67` is the canonical ID used across the CLI.
- `docs/INTEGRATION_TESTING.md` gains "Authentication helpers" and "Output capture and parallelism" sections covering the above.
- CI already runs `./internal/commands/ports/...` in `integration-test.yml`; the new file is auto-discovered.

Tests passing against Staging API on 2026/06/03

### DIFF
--- a/internal/commands/ports/ports_integration_test.go
+++ b/internal/commands/ports/ports_integration_test.go
@@ -136,11 +136,14 @@ func registerPortCleanup(t *testing.T, client *megaport.Client, uid string) {
 		require.NoError(t, delCmd.Flags().Set("now", "true"))
 		require.NoError(t, delCmd.Flags().Set("force", "true"))
 
+		var deleteErr error
 		_ = captureWithFormat("table", func() {
-			if err := DeletePort(delCmd, []string{uid}, true); err != nil {
-				t.Errorf("cleanup: failed to delete port %s: %v", uid, err)
-			}
+			deleteErr = DeletePort(delCmd, []string{uid}, true)
 		})
+		if deleteErr != nil {
+			t.Errorf("cleanup: failed to delete port %s: %v", uid, deleteErr)
+			return
+		}
 
 		getCmd := newGetPortCmd()
 		var getErr error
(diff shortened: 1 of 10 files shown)


internal/commands/ports/ports_integration_test.go
```diff
@@ -0,0 +1,421 @@
...
+		"marketPlaceVisibility": false,
+	}
+	buyJSON, err := json.Marshal(buyPayload)
+	require.NoError(t, err)
+
+	buyCmd := newBuyPortCmd()
+	require.NoError(t, buyCmd.Flags().Set("json", string(buyJSON)))
+
+	uid := runBuyPort(t, buyCmd, portName)
+	registerPortCleanup(t, uid)
```
projectsuperba: If the API creates a port but the UID extraction / the response fails, I think the cleanup of the resource will not occur

Phil-Browne: Fixed in 871ed20. `runBuyPort` and `runBuyLAGPort` now call a new `registerBuyCleanups` helper immediately after `BuyPort`/`BuyLAGPort` returns — before `uidFromBuyResponse` can call `t.FailNow`. The helper reads all UIDs from `integrationBuyResponses` (already populated by the `init` hook at that point) and registers a `t.Cleanup` for each one, so any created ports are deleted even if UID assertion subsequently fails.

The same pattern was applied to the MVE tests in #376 (e40e4f2): `t.Cleanup` is now registered before `require.NotEmpty(mveUID)`, with an early-return guard for the empty-UID edge case.

projectsuperba (approved):

PR #367, opened 2026-05-22
docs(ESD-1207): refresh docs and update alpha status

Phil-Browne:
Docs sweep against the current binary, plus a few small CLI fixes that surfaced during review.

## Docs

- Soften the root README's alpha caution to "alpha, beta coming soon".
- Regenerate `docs/` to cover nat-gateway, mcr looking-glass, and newer top-level flags (`--no-header`, `--no-pager`, `--template`, `go-template` output).
- Add the missing sections to the README command list (nat-gateway, product, topology, apply, and auth/whoami/config/completion/generate-docs/version).
- Fix `megaport` → `megaport-cli` and singular → plural command names (`port list` → `ports list`) across README, CONTRIBUTING, CHANGELOG, WASM_README, and the frontend-integration guides.
- WASM_README: list the modules actually compiled into the browser build, and call out the ones deliberately excluded.
- Bump CONTRIBUTING's prerequisite to Go 1.25 (matches `go.mod`); correct the config doc to `ConfigVersion = 1`.

## Small CLI changes (from review)

- `nat-gateway create` uses a new `DesignConfirmPrompt` ("Proceed with creation?") and a "Create cancelled" message instead of the billing-oriented `BuyConfirmPrompt` — it only produces a non-billable DESIGN-state record.
- It also gets its own `--yes`/`-y` flag whose help notes that no charges are incurred until `nat-gateway buy`, replacing the shared `WithBuyConfirmFlags()` whose wording implied a purchase.
- `mve buy`'s JSON example drops the real-looking `adminPassword` for a `<your-admin-password>` placeholder, with a note to treat credential fields as secrets.

## Test plan

- [ ] `go build` succeeds and `generate-docs ./docs` is idempotent against the committed tree
- [ ] `go test ./...` passes
- [ ] README, WASM_README, CONTRIBUTING render correctly on GitHub

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 [![codecov](https://codecov.io/gh/megaport/megaport-cli/branch/main/graph/badge.svg)](https://codecov.io/gh/megaport/megaport-cli)
 
-> [!CAUTION]
-> The Megaport CLI tool is currently an unsupported alpha, we're excited for feedback but please know that functionality and features may change drastically, and there may be bugs.
+> [!NOTE]
+> The Megaport CLI is currently in alpha, with a beta release coming soon. We're excited for feedback, but please know that functionality and features may still change and there may be bugs.
 
 ## Table of Contents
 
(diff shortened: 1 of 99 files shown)


docs/megaport-cli_mve_buy.md
```diff
@@ -24,6 +24,7 @@ This command allows you to purchase an MVE by providing the necessary details.
 
 ### Important Notes
   - For production deployments, you may want to use a JSON file to manage complex configurations
+  - Treat fields like adminPassword, adminSshPublicKey, sshPublicKey, accountKey, and any other credentials as secrets — replace the placeholders in these examples and avoid committing populated config files to source control.
```
mp1-tomclancy: **thought:** Public keys probably don't need to be included in this list of secrets. Obviously it will be case-by-case for users, but committing public keys doesn't carry the same risks as passwords/account keys etc.

README.md
```diff
@@ -244,12 +244,24 @@ Key architectural components:
...
+- `nat-gateway`: Manage NAT Gateways
 - `partners`: List and search partner ports
+- `product`: List products and look up product types by UID
 - `servicekeys`: Manage service keys
 - `users`: Manage company users
 - `managed-account`: Manage partner managed accounts (sub-companies)
 - `billing-market`: View and configure billing market details
 - `status`: Show a dashboard of all resources
+- `topology`: Show a tree of resources and their VXC connections
+- `apply`: Provision multiple resources from a declarative YAML or JSON config
```
mp1-tomclancy, line 256: **question:** is there a rule for which resources are plural vs. singular. e.g.:

`product` vs. `partners`

what do you think about standardising on one form, but then providing some aliases?

mp1-tomclancy (approved):
Looks good. Left some non-blocking comments for future discussion.

docs/megaport-cli_mve_buy.md
Phil-Browne: Fixed. Removed public key fields from the list — only adminPassword, accountKey, and other actual credentials remain.

PR #374, opened 2026-05-28
WIP ENG-31566: Allow portal to override hostname-derived environment in WASM

pedrotorchio-megaport:
## Contributions

Please read the (Contribution Guidelines)[https://github.com/megaport/megaportgo/wiki/Contributing.md]
prior to lodging Pull Requests (PR).

## Description

Adds an optional third `environment` argument to the WASM `setAuthToken` function so the portal can explicitly pin the backend environment instead of relying solely on hostname-based auto-detection.

Portal console
image

Portal UI
image

Different hostname and environment
image

### Motivation

`setAuthToken(token, hostname)` derives the environment and API URL purely from the hostname via `hostnameToEnvironment` / `hostnameToAPIURL`. That mapping has assumptions baked in that don't hold for every portal deployment — most notably that development environments (localhost and other non prod, such as qa, uat, etc, as well as ip addresses) don't map cleanly to the desired backend and there was no way for the caller to correct that mapping.

Changing the mapping itself would be a breaking change for any existing two-arg caller relying on the legacy behaviour, so this PR takes the opt-in approach: callers that know which environment they want can pass it as a third argument and bypass the hostname inference entirely.

### Behaviour

- Two-argument callers are unaffected — the hostname-derived environment/URL is still used.
- With a third argument:
  - `"production"` resolves to `https://api.megaport.com/`.
  - Any other value resolves to `https://api-<env>.megaport.com/`.
  - Input is normalised with `TrimSpace` + `ToLower`; whitespace-only is treated as no override.
- Console logs were restructured to clearly show the hostname, the resolved environment (with a `(derived from hostname)` or `(mismatch with provided hostname. Expected '…')` annotation), and the API URL, plus a status emoji (🟢 development / 🟠 production / 🔴 mismatch) so the chosen environment is obvious in the browser console.

### Breaking changes
  - WASM setAuthToken: unrecognized hostnames now fail closed; previously they silently
  defaulted to production.
  - WASM setAuthToken: localhost, 127.*, 192.168.*, 10.* no longer auto-map to development
   — callers must now pass an explicit environment override.
  - WASM setAuthToken: hostname-to-env resolution no longer uses substring matching for
  staging/dev/uat/qa; only the strict <app>-<env>.megaport.com pattern is honoured.
  - WASM setAuthToken: production hostname pattern now requires a single-word app name (no
   hyphens); hyphenated subdomains like my-app.megaport.com are no longer treated as
  production.

### Files

- `internal/wasm/wasm.go` — `setAuthToken` accepts the optional `environment` arg, restructures resolution and logging.
- `internal/wasm/wasm_test.go` — adds coverage for the override path (including the localhost-override scenario the change exists for), production override, mixed-case and whitespace normalisation, and the no-mismatch path.
- `frontend-integration/types/megaport-wasm.d.ts` — TypeScript bindings updated to expose the optional `environment` parameter and the `apiURL` field on the global shim's return type.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# Contributor Agreement

Lodging this Pull Request (PR) indicates agreement with the project's
(Contributor License Agreement)[https://github.com/megaport/megaportgo/wiki/Megaport_Contributor_Licence_Agreement.md].

Please read the Contributor Licence Agreement (CLA) and affirm your acceptance here:

[I have read an accept the CLA]

**NOTE** If multiple authors have commited to this PR, each one will need to comment on this PR and
agree to the CLA before this PR can be accepted.


internal/wasm/wasm.go
```diff
@@ -718,12 +721,49 @@ func setAuthToken(this js.Value, args []js.Value) interface{} {
...
+	var environment, apiURL string
+	if explicitEnv == "" {
+		environment = derivedEnv
+		apiURL = derivedURL
+	} else if explicitEnv == "production" {
+		environment = "production"
+		apiURL = "https://api.megaport.com/"
+	} else {
+		environment = explicitEnv
+		apiURL = fmt.Sprintf("https://api-%s.megaport.com/", explicitEnv)
```
Phil-Browne: Security: worth fixing before merge. `environment` is free text (line 727) and is interpolated directly into the API base URL the token is sent to, skipping the `*.megaport.com` guard the hostname path applies. For example, `environment = "evil.com/"` yields `https://api-evil.com/.megaport.com/` (host `api-evil.com`), and `loginFunc`'s `WithBaseURL` does not validate the host, so the token is sent there.

An allowlist closes it:

```go
var envNamePattern = regexp.MustCompile(`^[a-z0-9-]+$`)
if explicitEnv != "" && !envNamePattern.MatchString(explicitEnv) {
    return map[string]interface{}{"success": false, "error": "invalid environment override"}
}
```

internal/wasm/wasm.go
```diff
@@ -718,12 +721,49 @@ func setAuthToken(this js.Value, args []js.Value) interface{} {
...
+
+	// Derive environment and API URL from hostname, but allow explicit override for flexibility.
+	derivedEnv := hostnameToEnvironment(hostname)
+	derivedURL := hostnameToAPIURL(hostname)
+
+	var environment, apiURL string
+	if explicitEnv == "" {
+		environment = derivedEnv
+		apiURL = derivedURL
+	} else if explicitEnv == "production" {
```
Phil-Browne: Explicit `api-<env>` does not match the hostname path for non-prod envs. For example, `"development"` yields `api-development...` here, but the derived path gives `api-mpone-dev...` (`qa` and `uat` differ too). Only `"production"` is special-cased, so `"prod"` silently becomes `api-prod...`.

Routing both through a single `environmentToAPIURL(env)` helper would stop them drifting. Does `api-qa.megaport.com` exist, or should `qa` map to `mpone-dev` like the hostname path does?

internal/wasm/wasm.go
```diff
@@ -718,12 +721,49 @@ func setAuthToken(this js.Value, args []js.Value) interface{} {
...
+	if explicitEnv == "" {
+		environmentLog = fmt.Sprintf("%s (derived from hostname)", environment)
+	} else if explicitEnv != derivedEnv {
+		environmentLog = environment + fmt.Sprintf(" (mismatch with provided hostname. Expected '%s')", derivedEnv)
+	} else {
+		environmentLog = environment
+	}
+
+	var statusEmoji string
+	if explicitEnv != derivedEnv {
```
Phil-Browne: For a 2-arg call `explicitEnv` is `""`, so `"" != derivedEnv` is always true. Every no-override login is therefore flagged as a mismatch, even though the text on the same line says `(derived from hostname)`. Guard the mismatch branch on a real override:

```go
case explicitEnv != "" && explicitEnv != derivedEnv: // was: explicitEnv != derivedEnv
```

The tests do not capture `console.log` output, so CI does not catch this.

internal/wasm/wasm_test.go
```diff
@@ -522,57 +522,136 @@ func TestSetAuthToken(t *testing.T) {
 		name        string
 		token       string
 		hostname    string
+		explicitEnv string
 		expectError bool
 		expectedEnv string
+		expectedURL string
```
Phil-Browne: Worth adding a few cases: a malicious `environment` (e.g. `"evil.com/"`), explicit `"development"`/`"qa"` to pin the URL, and an unknown like `"prod"`. Happy-path + whitespace/uppercase coverage is great.

Phil-Browne (commented):
Thanks for this. The opt-in approach (rather than changing the hostname mapping) is the right call, and the test table is well structured. Keeping the token out of the JS globals and exposing only metadata is a nice touch.

I left a few inline notes. The main one is a security concern on how the override builds the API URL, which I would suggest tightening before merge, since that URL is where the bearer token is sent. The others are a small logic bug in the new console logging and a couple of consistency points on the environment to URL mapping.

Happy to talk through any of it.

internal/wasm/wasm.go
```diff
@@ -718,12 +721,49 @@ func setAuthToken(this js.Value, args []js.Value) interface{} {
...
+	if explicitEnv == "" {
+		environment = derivedEnv
+		apiURL = derivedURL
+	} else if explicitEnv == "production" {
+		environment = "production"
+		apiURL = "https://api.megaport.com/"
+	} else {
+		environment = explicitEnv
+		apiURL = fmt.Sprintf("https://api-%s.megaport.com/", explicitEnv)
+	}
```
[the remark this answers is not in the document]
pedrotorchio-megaport: good point. Fixing it.

internal/wasm/wasm.go
```diff
@@ -718,12 +721,49 @@ func setAuthToken(this js.Value, args []js.Value) interface{} {
...
+
+	var statusEmoji string
+	if explicitEnv != derivedEnv {
+		statusEmoji = "🔴"
+	} else if environment == "production" {
+		statusEmoji = "🟠"
+	} else {
+		statusEmoji = "🟢"
+	}
```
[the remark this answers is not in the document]
pedrotorchio-megaport: the logic here is just wrong, I'm rewriting it

internal/wasm/wasm.go
pedrotorchio-megaport: yeah I fixed this one already. I should have this PR as WIP (wasn't aware it would notify you immediately 😅 )

internal/wasm/wasm_test.go
pedrotorchio-megaport: done with the fix

Phil-Browne (approved):

Phil-Browne (commented):
A few review notes. One is on `frontend-integration/composables/useMegaportWASM.ts:461`, which isn't part of this PR's diff so it can't be left inline:

This composable is the portal's actual entry point, and it still has the two-arg signature `setAuthToken(token, hostname?)`, it never forwards the new `environment` override, so the third argument added in this PR isn't reachable from the portal.

It would also regress local dev: `actualHostname` falls back to `window.location.hostname` (`localhost`), and with this PR `setAuthToken(token, "localhost")` now fails closed instead of mapping to development. So local development through this composable breaks until it's updated.

I would suggest adding an `environment?` param here and plumbing it through to `window.setAuthToken(token, actualHostname, environment)` (with a sensible localhost default) as part of this PR.

internal/wasm/wasm.go
```diff
@@ -16,6 +17,83 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const API_URL_PRODUCTION = "https://api.megaport.com/"
+const API_URL_NON_PRODUCTION_TEMPLATE = "https://api-%s.megaport.com/"
```
Phil-Browne: `API_URL_PRODUCTION` / `API_URL_NON_PRODUCTION_TEMPLATE` are all caps snake-case. The convention for `megaportgo` and `megaport-cli` is `mixedCaps`. I would suggest `apiURLProduction` and `apiURLNonProductionTemplate`.

internal/wasm/wasm.go
```diff
@@ -718,17 +801,90 @@ func setAuthToken(this js.Value, args []js.Value) interface{} {
...
+	// In either case we want to flag this clearly in the console logs so it's obvious to the user and can be easily debugged.
+	mismatch := explicitEnv != "" && derivedOK && explicitEnv != derivedEnv
+
+	// Build a console-friendly description of the resolved environment, annotating
+	// where it came from and flagging override-vs-hostname mismatches.
+	var environmentLog string
+	if explicitEnv == "" {
+		environmentLog = fmt.Sprintf("%s (derived from hostname)", environment)
+	} else if mismatch {
+		environmentLog = environment + fmt.Sprintf(" (mismatch with hostname-derived %q)", derivedEnv)
```
Phil-Browne: Minor: this branch concatenates then `Sprintf`s, while the sibling branch just above uses a single `fmt.Sprintf`. Fold this into one `fmt.Sprintf` for consistency.

internal/wasm/wasm.go
```diff
@@ -16,6 +17,83 @@ import (
...
+		return m[1], true
+	}
+	return "", false
+}
+
+// restrictEnvironmentName collapses an environment name into one of the three
+// canonical values that downstream consumers of MEGAPORT_ENVIRONMENT understand
+// ("production" / "staging" / "development"). Anything that isn't "production"
+// or "staging" collapses to "development".
+func restrictEnvironmentName(env string) string {
```
Phil-Browne, line 100: This sits alongside `normalizeEnvironment` in `config_shared.go`, and the two disagree on `"prod"` (here it buckets to development; `normalizeEnvironment` maps it to production). The divergence looks intentional and the tests document it, but two similarly named env normalizers that disagree could present a future foot-gun. A one-line doc note cross-referencing the other and saying why they differ would help.

(Not a routing bug: the token path in `login_wasm.go` prefers `apiURL` via `WithBaseURL`, so the bucketed `MEGAPORT_ENVIRONMENT` doesn't affect where the token routes.)

PR #376, opened 2026-06-03
ESD-1309: Migrate MCR and MVE integration tests to direct action function calls

Phil-Browne:
Migrates the MCR and MVE integration tests from Approach A (driving the compiled binary via `exec.Command`) to Approach B: tests live in each resource package, gated behind `//go:build integration`, and call the action functions directly with a mock-injected staging client.

- Add `internal/commands/mcr/mcr_integration_test.go` — full MCR lifecycle (flag-based and JSON input), plus the prefix filter list create/get/update/delete lifecycle, with cleanup ordered so the filter list is removed before the MCR.
- Add `internal/commands/mve/mve_integration_test.go` — full MVE lifecycle (flag-based and JSON input). Image IDs are discovered dynamically via `ListMVEImages` since they change on staging; no hardcoded image IDs.
- Both files use `testutil.SetupIntegrationClient` + `testutil.LoginWithClient`, lean on the SDK's wait-for-provision so there are no `time.Sleep` calls, and generate unique resource names per run.

Also fixes MVE validation and integration test logic for MVE Size. 

This is a follow up to https://github.com/megaport/megaport-cli/pull/344 allowing for Integration Tests to operate across a CI pipeline. 

MCR and MVE integration tests passing against staging 2026/06/03

### DIFF
--- a/internal/commands/mcr/mcr_integration_test.go
+++ b/internal/commands/mcr/mcr_integration_test.go
@@ -157,9 +157,12 @@ func TestIntegration_MCRLifecycle(t *testing.T) {
 	require.NoError(t, buyErr, "buy MCR output: %s", buyOut)
 
 	mcrUID := parseCreatedUID(buyOut, "MCR")
-	require.NotEmpty(t, mcrUID, "could not parse MCR UID from: %s", buyOut)
-
+	// Register cleanup before asserting on mcrUID, so any created MCR is
+	// deleted even if the UID parse fails.
 	t.Cleanup(func() {
+		if mcrUID == "" {
+			return
+		}
 		delCmd := integrationMCRDeleteCmd()
 		_ = delCmd.Flags().Set("force", "true")
 		var delErr error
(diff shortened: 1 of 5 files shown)


internal/commands/mcr/mcr_integration_test.go
```diff
@@ -0,0 +1,330 @@
...
+	require.NoError(t, buyCmd.Flags().Set("yes", "true"))
+
+	var buyErr error
+	buyOut := captureTableOutput(func() { buyErr = BuyMCR(buyCmd, nil, true) })
+	require.NoError(t, buyErr, "buy MCR output: %s", buyOut)
+
+	mcrUID := parseCreatedUID(buyOut, "MCR")
+	require.NotEmpty(t, mcrUID, "could not parse MCR UID from: %s", buyOut)
+
+	t.Cleanup(func() {
```
projectsuperba, line 162: Can the ordering of cleanup registration be handled the same way MVE is to prevent leaked resources?

Phil-Browne: Done — both MCR tests (flag-based and JSON input) now register the cleanup before asserting on the UID, matching the MVE pattern with a nil guard in the closure. Pushed in 4682b9c.

projectsuperba (approved):

PR #413, opened 2026-06-15
test(mcr): add IPSec add-on lifecycle integration test

Phil-Browne:
Extends the MCR provisioning integration tests with an IPSec add-on lifecycle, closing the gap where the IPSec add-on add and update commands (including the disable path) were never exercised against staging.

The test provisions an MCR, adds an IPSec add-on, polls until the add-on UID surfaces, updates the tunnel count, then disables it (tunnel-count 0). The MCR is torn down in `t.Cleanup`. It runs only under the `integration provisioning` tags (manual job) and skips cleanly when staging reports the add-on is unsupported on the test MCR, so a real request-building or SDK regression still fails the test rather than being masked.

Integration tests pass against staging: the full add, update, then disable lifecycle runs end to end (~47s) with the MCR cleaned up afterward.

Ticket: ESD-1391

### DIFF
--- a/internal/commands/mcr/mcr_ipsec_addon_lifecycle_integration_test.go
+++ b/internal/commands/mcr/mcr_ipsec_addon_lifecycle_integration_test.go
@@ -1,0 +1,202 @@
+//go:build integration && provisioning
+
+package mcr
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/megaport/megaport-cli/internal/base/output"
+	"github.com/megaport/megaport-cli/internal/testutil"
+	megaport "github.com/megaport/megaportgo"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


penzeliz-megaport (approved):
LGTM. Well-structured IPSec add-on lifecycle test. Correctly tagged `//go:build integration && provisioning` since it provisions and bills a real MCR. The add-on UID comes from an SDK read (`GetMCR` → `mcr.AddOns[].AddOnUID`), not output scraping, and the disable assertion targets a genuinely distinct source branch (`"IPSec add-on disabled successfully"` fires only when `tunnelCount == 0`), so add/update/disable are each really exercised. Cleanup is sound: `t.Cleanup` for MCR delete is registered right after buy (before the UID assertion), and login-restore correctly uses `t.Cleanup` rather than `defer` so it can't swap environments before the delete runs. Polls with a 15s/5min budget instead of fixed sleeps, and `isIPSecAddOnUnavailable` is narrowly scoped (`not supported`/`unsupported`/`not eligible`, explicitly excluding auth/transient wording) so a real regression fails rather than skips. Approving.

PR #414, opened 2026-06-15
test(ESD-1392): resource tag lifecycle integration tests (ports/mcr/mve/vxc)

Phil-Browne:
Adds an update-tags then list-tags round-trip to the existing provisioning lifecycle tests for ports, MCR, MVE, and VXC. Each step sets two tags on the resource the lifecycle test already provisioned, reads them back, asserts they round-trip, then clears them. It rides on the existing resource, so there's no new cleanup and little added runtime.

- ports/vxc (parallel tests): exercise the update-tags command, then read tags back via the SDK for assertions. These tests can't use `CaptureOutput` under `t.Parallel()` (it swaps global stdout and races with concurrent spinners), matching the existing `portFromSDK`/`vxcFromSDK` pattern.
- mcr/mve (serial tests): exercise both the update-tags and list-tags commands, capturing the list-tags JSON and asserting it contains the tags, then clearing and asserting they're gone.
- Set/clear both pass `--force` so the confirm prompt never blocks on stdin in CI.
- Round-trip assertions check our tags are present/absent rather than requiring an exact whole-map match, so an API-injected tag can't cause flakiness.

Runs in the manual provisioning job under the `integration provisioning` build tags.

Verified by running the full `integration provisioning` suite against staging: all four lifecycle tests (Port, VXC, MCR, MVE) pass, provisioning and tearing down real resources cleanly with no failures.

### DIFF
--- a/internal/commands/mcr/mcr_integration_test.go
+++ b/internal/commands/mcr/mcr_integration_test.go
@@ -76,6 +76,15 @@ func integrationMCRDeleteCmd() *cobra.Command {
 	return cmd
 }
 
+func integrationMCRTagCmd() *cobra.Command {
+	cmd := &cobra.Command{Use: "update-tags"}
+	cmd.Flags().Bool("interactive", false, "")
+	cmd.Flags().BoolP("force", "f", false, "")
+	cmd.Flags().String("json", "", "")
+	cmd.Flags().String("json-file", "", "")
+	return cmd
+}
+
 func integrationMCRPrefixFilterCmd() *cobra.Command {
 	cmd := &cobra.Command{Use: "prefix-filter-list"}
 	cmd.Flags().Bool("interactive", false, "")
@@ -191,6 +200,40 @@ func TestIntegration_MCRLifecycle(t *testing.T) {
 	require.NoError(t, updErr, "update MCR output: %s", updOut)
 
 	assert.Equal(t, newName, getMCRJSON(t, mcrUID)["name"])
+
+	// Resource tag round-trip (ESD-1392): set tags via update-tags, read them
+	// back via list-tags, then clear them. Rides on the lifecycle MCR, so no
+	// extra cleanup is needed.
+	setTagsCmd := integrationMCRTagCmd()
+	require.NoError(t, setTagsCmd.Flags().Set("json", `{"env":"cli-integration","owner":"esd-1392"}`))
+	require.NoError(t, setTagsCmd.Flags().Set("force", "true"))
+	var setTagsErr error
+	setTagsOut := captureTableOutput(func() { setTagsErr = UpdateMCRResourceTags(setTagsCmd, []string{mcrUID}, true) })
+	require.NoError(t, setTagsErr, "update MCR tags output: %s", setTagsOut)
+
+	var listTagsErr error
+	listTagsOut := output.CaptureOutput(func() {
+		listTagsErr = ListMCRResourceTags(&cobra.Command{Use: "list-tags"}, []string{mcrUID}, true, "json")
+	})
+	require.NoError(t, listTagsErr, "list MCR tags output: %s", listTagsOut)
+	assert.Contains(t, listTagsOut, "cli-integration")
+	assert.Contains(t, listTagsOut, "esd-1392")
+
+	// Clear the tags so the MCR is left clean for the steps that follow.
+	clearTagsCmd := integrationMCRTagCmd()
+	require.NoError(t, clearTagsCmd.Flags().Set("json", "{}"))
+	require.NoError(t, clearTagsCmd.Flags().Set("force", "true"))
+	var clearTagsErr error
+	clearTagsOut := captureTableOutput(func() { clearTagsErr = UpdateMCRResourceTags(clearTagsCmd, []string{mcrUID}, true) })
+	require.NoError(t, clearTagsErr, "clear MCR tags output: %s", clearTagsOut)
+
+	var verifyTagsErr error
+	verifyTagsOut := output.CaptureOutput(func() {
+		verifyTagsErr = ListMCRResourceTags(&cobra.Command{Use: "list-tags"}, []string{mcrUID}, true, "json")
+	})
+	require.NoError(t, verifyTagsErr, "list MCR tags after clear output: %s", verifyTagsOut)
+	assert.NotContains(t, verifyTagsOut, "cli-integration")
+	assert.NotContains(t, verifyTagsOut, "esd-1392")
 
 	// Prefix filter list lifecycle (create -> get -> update -> delete).
 	createPFLJSON := `{
--- a/internal/commands/mcr/mcr_integration_test.go
+++ b/internal/commands/mcr/mcr_integration_test.go
@@ -85,6 +85,20 @@ func integrationMCRTagCmd() *cobra.Command {
 	return cmd
 }
 
+// tagsFromListJSON parses the JSON output of list-tags (an array of {key,value}
+// objects) into a map, so assertions can check key->value pairs instead of
+// substring-matching the rendered blob.
+func tagsFromListJSON(t *testing.T, out string) map[string]string {
+	t.Helper()
+	var tags []output.ResourceTag
+	require.NoErrorf(t, json.Unmarshal([]byte(out), &tags), "parse list-tags JSON: %s", out)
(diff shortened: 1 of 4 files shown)


internal/commands/mcr/mcr_integration_test.go
```diff
@@ -192,6 +201,40 @@ func TestIntegration_MCRLifecycle(t *testing.T) {
...
+	setTagsOut := captureTableOutput(func() { setTagsErr = UpdateMCRResourceTags(setTagsCmd, []string{mcrUID}, true) })
+	require.NoError(t, setTagsErr, "update MCR tags output: %s", setTagsOut)
+
+	var listTagsErr error
+	listTagsOut := output.CaptureOutput(func() {
+		listTagsErr = ListMCRResourceTags(&cobra.Command{Use: "list-tags"}, []string{mcrUID}, true, "json")
+	})
+	require.NoError(t, listTagsErr, "list MCR tags output: %s", listTagsOut)
+	assert.Contains(t, listTagsOut, "cli-integration")
+	assert.Contains(t, listTagsOut, "esd-1392")
```
projectsuperba: Optional: assert key→value pairs instead of substring-matching the blob (matches the ports/VXC SDK assertions). Needs the `tagsFromListJSON` helper from the review summary.

```suggestion
	require.NoError(t, listTagsErr, "list MCR tags output: %s", listTagsOut)
	want := map[string]string{"env": "cli-integration", "owner": "esd-1392"}
	got := tagsFromListJSON(t, listTagsOut)
	for k, v := range want {
		assert.Equalf(t, v, got[k], "tag %q should round-trip", k)
	}
```

internal/commands/mcr/mcr_integration_test.go
```diff
@@ -192,6 +201,40 @@ func TestIntegration_MCRLifecycle(t *testing.T) {
...
+	clearTagsOut := captureTableOutput(func() { clearTagsErr = UpdateMCRResourceTags(clearTagsCmd, []string{mcrUID}, true) })
+	require.NoError(t, clearTagsErr, "clear MCR tags output: %s", clearTagsOut)
+
+	var verifyTagsErr error
+	verifyTagsOut := output.CaptureOutput(func() {
+		verifyTagsErr = ListMCRResourceTags(&cobra.Command{Use: "list-tags"}, []string{mcrUID}, true, "json")
+	})
+	require.NoError(t, verifyTagsErr, "list MCR tags after clear output: %s", verifyTagsOut)
+	assert.NotContains(t, verifyTagsOut, "cli-integration")
+	assert.NotContains(t, verifyTagsOut, "esd-1392")
```
projectsuperba: Optional: key-based clear assertion (reuses `want` from the set step above; won't pass trivially on empty output).

```suggestion
	require.NoError(t, verifyTagsErr, "list MCR tags after clear output: %s", verifyTagsOut)
	cleared := tagsFromListJSON(t, verifyTagsOut)
	for k := range want {
		assert.NotContainsf(t, cleared, k, "tag %q should be cleared", k)
	}
```

internal/commands/mve/mve_integration_test.go
```diff
@@ -228,6 +237,40 @@ func TestIntegration_MVELifecycle(t *testing.T) {
...
+	setTagsOut := captureTableOutput(func() { setTagsErr = UpdateMVEResourceTags(setTagsCmd, []string{mveUID}, true) })
+	require.NoError(t, setTagsErr, "update MVE tags output: %s", setTagsOut)
+
+	var listTagsErr error
+	listTagsOut := output.CaptureOutput(func() {
+		listTagsErr = ListMVEResourceTags(&cobra.Command{Use: "list-tags"}, []string{mveUID}, true, "json")
+	})
+	require.NoError(t, listTagsErr, "list MVE tags output: %s", listTagsOut)
+	assert.Contains(t, listTagsOut, "cli-integration")
+	assert.Contains(t, listTagsOut, "esd-1392")
```
projectsuperba: Optional: assert key→value pairs instead of substring-matching the blob (matches the ports/VXC SDK assertions). Needs the `tagsFromListJSON` helper from the review summary.

```suggestion
	require.NoError(t, listTagsErr, "list MVE tags output: %s", listTagsOut)
	want := map[string]string{"env": "cli-integration", "owner": "esd-1392"}
	got := tagsFromListJSON(t, listTagsOut)
	for k, v := range want {
		assert.Equalf(t, v, got[k], "tag %q should round-trip", k)
	}
```

internal/commands/mve/mve_integration_test.go
```diff
@@ -228,6 +237,40 @@ func TestIntegration_MVELifecycle(t *testing.T) {
...
+	clearTagsOut := captureTableOutput(func() { clearTagsErr = UpdateMVEResourceTags(clearTagsCmd, []string{mveUID}, true) })
+	require.NoError(t, clearTagsErr, "clear MVE tags output: %s", clearTagsOut)
+
+	var verifyTagsErr error
+	verifyTagsOut := output.CaptureOutput(func() {
+		verifyTagsErr = ListMVEResourceTags(&cobra.Command{Use: "list-tags"}, []string{mveUID}, true, "json")
+	})
+	require.NoError(t, verifyTagsErr, "list MVE tags after clear output: %s", verifyTagsOut)
+	assert.NotContains(t, verifyTagsOut, "cli-integration")
+	assert.NotContains(t, verifyTagsOut, "esd-1392")
```
projectsuperba: Optional: key-based clear assertion (reuses `want` from the set step above; won't pass trivially on empty output).

```suggestion
	require.NoError(t, verifyTagsErr, "list MVE tags after clear output: %s", verifyTagsOut)
	cleared := tagsFromListJSON(t, verifyTagsOut)
	for k := range want {
		assert.NotContainsf(t, cleared, k, "tag %q should be cleared", k)
	}
```

projectsuperba (commented):
I had a go at this review with Claude assistance to also generate code suggestions. Let me know if that is useful to help get these PRs turned around quicker!

Some very minor strengthening suggestions.

The MCR/MVE checks use `assert.Contains`/`NotContains` on the raw JSON blob — that verifies the two *values* appear but not the keys or their pairing, and `NotContains` can pass trivially on empty output. Parsing the `list-tags` JSON into a map brings them to the same rigor as the ports/VXC SDK assertions while still exercising the command. The inline suggestions depend on this helper, added to both `mcr_integration_test.go` and `mve_integration_test.go` (`encoding/json` and `output` are already imported in both):

```go
// tagsFromListJSON unmarshals the list-tags JSON (an array of {"key","value"})
// into a map for key→value assertions.
func tagsFromListJSON(t *testing.T, out string) map[string]string {
	t.Helper()
	var tags []output.ResourceTag
	require.NoErrorf(t, json.Unmarshal([]byte(out), &tags), "parse list-tags JSON: %s", out)
	m := make(map[string]string, len(tags))
	for _, tag := range tags {
		m[tag.Key] = tag.Value
	}
	return m
}
```

internal/commands/mcr/mcr_integration_test.go
Phil-Browne: Done in ea30486. Added a tagsFromListJSON helper (parses the list-tags []ResourceTag JSON into a map) and assert key->value round-trip. Also switched the set step to json.Marshal(want) so set and assert share one source of truth, matching the ports/VXC pattern.

internal/commands/mve/mve_integration_test.go
Phil-Browne: Done in ea30486. Added a tagsFromListJSON helper (parses the list-tags []ResourceTag JSON into a map) and assert key->value round-trip. Also switched the set step to json.Marshal(want) so set and assert share one source of truth, matching the ports/VXC pattern.

internal/commands/mcr/mcr_integration_test.go
Phil-Browne: Done in ea30486. Clear step now parses via tagsFromListJSON and asserts NotContains by key, so it no longer passes trivially on empty output.

internal/commands/mve/mve_integration_test.go
Phil-Browne: Done in ea30486. Clear step now parses via tagsFromListJSON and asserts NotContains by key, so it no longer passes trivially on empty output.

projectsuperba (approved):

PR #415, opened 2026-06-15
test(ESD-1393): add MCR/vrouter + inner-VLAN VXC provisioning lifecycle test

Phil-Browne:
Adds a provisioning lifecycle integration test for the MCR-attached (vrouter A-End) and inner-VLAN (Q-in-Q) VXC paths, which the existing port-to-port lifecycle tests never reach. A regression in `vxc_inputs_partner.go` (vrouter config building) or the inner-VLAN build path would currently go uncaught.

What's here:
- `TestIntegration_VXCMCRVrouterInnerVLANLifecycle`: buys an MCR (A-End vrouter partner config), a port (B-End with outer + inner VLAN), provisions the VXC, asserts the VLANs, and tears everything down via `t.Cleanup`. All resources are `CLI-Test-` prefixed.
- A small integration-only buy hook on the `mcr` package (mirrors the existing `ports` hook) so the test recovers the MCR UID from the SDK response instead of scraping stdout.
- Inner VLAN sits on the port B-End because Q-in-Q is a port handoff feature; VLAN/inner-VLAN are polled before asserting since those fields can lag the LIVE transition.

Cloud partner VXCs (AWS/Azure/Google/Oracle/IBM) are explicitly out of scope: they need live cloud-provider accounts on staging and should be tracked separately.

Verified against staging under `-tags 'integration provisioning'`: the new test passed on two consecutive runs (full provision, assert, and teardown each time, no leaked resources), alongside a clean run of the rest of the provisioning suite. Builds, vets, and lints clean. The provisioning run is manual-only (needs staging credentials).


penzeliz-megaport (approved):
Adds an MCR/vrouter + inner-VLAN VXC provisioning lifecycle test. The only non-`_test.go` file, `mcr_integration_hook.go`, is `//go:build integration`, so it never compiles into the shipped CLI binary. It's a clean test-support hook that records `BuyMCRResponse` UIDs so the VXC test recovers the MCR UID from the SDK response rather than scraping stdout. Effectively test-only, no production impact. CI green. LGTM.

PR #416, opened 2026-06-15
ESD-1491: route AWS/BGP ASN range checks through ValidateASN

Phil-Browne:
`ValidateAWSPartnerConfig` and `ValidateBGPConnectionConfig` previously hand-rolled an ASN range check against the raw `MaxASN` constant. On 32-bit targets (js/wasm), `ValidateASN` caps the upper bound at `math.MaxInt32` via `maxSupportedASNForInt()`, but the two validators bypassed that by comparing against `MaxASN` directly.

Both validators now call `ValidateASN` for the range check and re-wrap any error to preserve their existing field names and message format. `maxSupportedASNForInt()` is the single source of truth for the platform-aware bound.

- `ValidateAWSPartnerConfig`: replaces direct `> MaxASN` comparison with `ValidateASN` delegation
- `ValidateBGPConnectionConfig`: same
- Adds `TestValidateAWSPartnerConfig_HighASN` mirroring the existing BGP high-ASN boundary test

No behavior change on 64-bit targets. Error message wording is unchanged.

### DIFF
--- a/internal/validation/vxc.go
+++ b/internal/validation/vxc.go
@@ -236,8 +236,8 @@ func ValidateAWSPartnerConfig(config *megaport.VXCPartnerConfigAWS) error {
 	if config.ASN == 0 {
 		return NewValidationError("ASN", config.ASN, "cannot be empty")
 	}
-	if int64(config.ASN) < MinASN || int64(config.ASN) > MaxASN {
-		return NewValidationError("ASN", config.ASN, fmt.Sprintf("must be between %d-%d", MinASN, MaxASN))
+	if err := ValidateASN(config.ASN); err != nil {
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):
Routes AWS and BGP ASN range checks through the central `ValidateASN` instead of inlining `MinASN`/`MaxASN`. On 64-bit the bound is unchanged (`MaxASN` == `maxSupportedASNForInt()` == 4294967295); on 32-bit/wasm it correctly caps to int-max rather than overflowing. The re-wrap preserves the field name and message format, and the new `TestValidateAWSPartnerConfig_HighASN` pins the boundary. CI green. LGTM.

PR #417, opened 2026-06-16
test: add native-binary e2e harness and smoke test (ESD-1493)

Phil-Browne:
Stands up the `e2e/` package and a small harness that runs the compiled CLI as a real subprocess, so we finally exercise argv parsing through to exit code and stdout/stderr. Every other test today is in-process via `config.LoginFunc`. This unblocks the hermetic suite and the staging suite, both incoming follow-up PRs, which only add test files on top.

What's here:
- New `e2e/` package, all files behind `//go:build e2e`, so `go test ./...` and the production build ignore it.
- `TestMain` builds the CLI once (or uses a prebuilt binary when `MEGAPORT_CLI_E2E_BIN` is set), from the module root.
- `Run` execs the binary in a sandboxed env (isolated `HOME`, minimal `PATH`), capturing stdout, stderr, and exit code separately, with a 30s timeout. `RunWithEnv` opt-in forwards `MEGAPORT_*` for the staging tier.
- Makefile: split the placeholder `e2e` target into hermetic `e2e` and `e2e-staging`.
- One `TestE2E_Smoke` test runs `version` and asserts exit 0 with non-empty stdout.

`make e2e` passes locally; `go test ./...`, `go build ./...`, and `golangci-lint run` are all clean.


penzeliz-megaport (approved):
Solid foundation for native-binary e2e testing. Test-only, no production code, and the whole `e2e/` package is behind `//go:build e2e` so the production build and `go test ./...` ignore it. The harness is well-built: build-once in `TestMain` (with `GOWORK=off` for reproducibility, or a validated prebuilt binary via `MEGAPORT_CLI_E2E_BIN`), a hermetic sandbox (isolated `HOME`, minimal `PATH`, all `MEGAPORT_*` dropped unless explicitly passed through), and correct timeout handling: it checks `ctx.Err() == DeadlineExceeded` before unwrapping the exit code, so a timeout-killed process isn't misreported as a real exit status.

Verified `go build -o bin .` at the module root is correct: `package main` lives in the repo-root `main.go`.

Two non-blocking caveats for follow-up:

1. **CI doesn't exercise this harness yet.** The `e2e` build tag excludes these files from the `Unit Tests`/`Lint` checks, and this PR adds no workflow step to run `make e2e`. So the green checks don't prove the harness builds or runs; only the local `make e2e` does. The harness can rot undetected until the follow-up suites wire it into CI. Worth a CI step (even just the hermetic `make e2e`) sooner rather than later.

2. **Unix-only.** `sandboxEnv` hardcodes `PATH=/usr/bin:/bin`, so this won't run on Windows. The comment scopes it to the darwin/linux targets deliberately; just flagging in case Windows e2e coverage is ever wanted.

Neither blocks. LGTM.

PR #421, opened 2026-06-16
feat: register nat-gateway module in WASM build

Phil-Browne:
Registers the `nat-gateway` module in the WASM build so the full NAT Gateway command set is available in the browser, matching the native CLI.

The module already uses the standard overridable func-pointer pattern calling the SDK directly, with no filesystem, `os/exec`, or build-tag dependencies, so this is register-and-go. No `nat_gateway_actions_wasm.go` override was needed.

- Add `nat_gateway.NewModule()` to `registerModules()` in `modules_wasm.go`.
- Add a `js,wasm` smoke test asserting `nat-gateway`, `list`, `get`, and `list-sessions` resolve in the WASM command tree.

Verified against Staging via the native binary (same SDK func-pointer paths the WASM build uses): `list`, `get`, `list-sessions` return real data; `telemetry` renders both json and table (no samples on the test gateway); `validate`/`create`/`buy`/`update`/`delete` are wired. The js/wasm build compiles clean.

ESD-1284

### DIFF
--- a/cmd/megaport/modules_wasm.go
+++ b/cmd/megaport/modules_wasm.go
@@ -6,6 +6,7 @@ import (
 	"github.com/megaport/megaport-cli/internal/commands/locations"
 	"github.com/megaport/megaport-cli/internal/commands/mcr"
 	"github.com/megaport/megaport-cli/internal/commands/mve"
+	"github.com/megaport/megaport-cli/internal/commands/nat_gateway"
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):
Registers the nat-gateway module in the WASM build, following the established pattern. The WASM-tagged registration test is well-scoped: `--help` only (no network), asserts the usage path, and checks for "unknown command" so an unregistered command would fail it. CI green. LGTM.

PR #422, opened 2026-06-16
feat: register ix module in WASM browser build (ESD-1285)

Phil-Browne:
Registers the `ix` module in the WASM browser build so IX commands are available in-browser, matching the native CLI.

- Add `ix.NewModule()` to `registerModules()` in `cmd/megaport/modules_wasm.go`.
- The `ix` module calls the SDK directly through the WASM fetch transport (no filesystem or `os/exec` deps), so it registers like `partners` with no per-command WASM override needed.
- Builds clean for `js/wasm` and native; `ix` package tests, `cmd` tests, and `go vet` all pass.

Browser-against-Staging verification of each command (`get`/`buy`/`update`/`delete`/`list`/`status`/`validate`) still needs a manual pass with a Staging session. A read-only smoke test (`TestIntegration_IXListAndGet`) already covers `list`+`get`. `buy` is mutating, so confirm it's safe in the Staging demo before surfacing it.

ESD-1285


penzeliz-megaport (dismissed):
Registers the ix module in the WASM build (one import, one `moduleRegistry.Register` line). Trivial and additive, CI green. LGTM. Minor, non-blocking: unlike its sibling #421 (nat-gateway), this doesn't add a WASM registration test for ix; worth a follow-up to keep coverage symmetric.

penzeliz-megaport (dismissed):

penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. `ix` module registration mirrors the native CLI and the module is SDK-direct (WASM-safe). Help-only smoke test covers all 8 subcommands with a `NotContains "unknown command"` guard. Minor nit (non-blocking): the usage anchor `"megaport-cli ix"` is a looser substring match than the `Usage:\n  megaport-cli ix` block anchor used in #441; the `NotContains` assertion compensates. Same `modules_wasm.go` rebase coordination with #441. Approving.

PR #424, opened 2026-06-16
fix: return errors on failed mutations and fix usage exit code (ESD-1497)

Phil-Browne:
Fixes two exit-code correctness bugs so scripts can trust the CLI's exit status.

**Failed mutations now return an error.** Several manage commands printed a "not successful" message but still returned `nil`, so a mutation the API reported as *not* successful exited 0. Every not-successful branch now returns a non-nil error (and so exits non-zero). Affected sites:

- ports: delete, restore, lock, unlock
- mcr: delete, restore; prefix-filter-list update and delete
- mve: delete

**Usage errors now exit 2.** `isCobraUsageError` didn't recognize cobra's arg-count message ("accepts N arg(s), received M"), so e.g. `ports get` with no args exited 1 instead of the documented usage code 2. Matching on the shared `arg(s)` substring covers all four cobra arg validators (ExactArgs, MinimumNArgs, MaximumNArgs, RangeArgs).

Tests assert both the error return and the usage exit code, and the not-successful cases now check the error carries the resource UID/ID.

Notes:
- `servicekeys` is intentionally untouched: ESD-1417 (#407) is in flight over the same file.
- The existing `PrintError` + `return err` pattern (double-print under the root's `SilenceErrors: false`) is left as-is. It's a codebase-wide convention; changing it only here would diverge, and a global rework is out of scope for an exit-code fix.

### DIFF
--- a/cmd/megaport/megaport.go
+++ b/cmd/megaport/megaport.go
@@ -313,9 +313,9 @@ func isCobraUsageError(msg string) bool {
 		"unknown command",
 		"unknown flag",
 		"unknown shorthand flag",
-		"accepts between",
-		"accepts at most",
-		"accepts at least",
+		// Covers every cobra arg-count validator: ExactArgs, MinimumNArgs,
+		// MaximumNArgs, and RangeArgs all phrase their error as "... arg(s) ...".
+		"arg(s)",
 		"required flag(s)",
 	}
 	for _, p := range cobraPatterns {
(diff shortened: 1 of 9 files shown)


penzeliz-megaport (approved):
Approving as the canonical implementation of the exit-code fix: usage errors exit 2, and failed mutations return an error instead of a silent warning. Cleaner and more complete than the duplicate #420: a single `arg(s)` matcher covering all cobra arg-count validators (with a comment), early-return guard clauses, and the resource UID included in each failure message, plus it carries the ticket (ESD-1497). CI green, mergeable.

Note: #420 makes the same fix across the same files and edits the same lines, so the two can't both merge. I've flagged #420 as superseded and suggested closing it. LGTM.

PR #425, opened 2026-06-16
fix(ESD-1500): stop retrying buy/order calls on ambiguous errors

Phil-Browne:
`utils.WithRetry` retries on 502/504, which a gateway can return *after* an order was already accepted upstream. Buy/order calls were wrapped in it, so an ambiguous gateway error could re-submit an accepted order and double-bill.

This adds `WithOrderRetry`, which retries only when the request provably never reached processing (HTTP 429 and connection-refused-before-send), and moves every non-idempotent create/buy path onto it.

- New `WithOrderRetry` helper: retries on 429 and pre-send connection-refused only, never on 5xx or other ambiguous errors.
- Switched all non-idempotent creates/buys: port, LAG port, VXC, MCR, MVE, NAT gateway create/buy, IX buy, MCR prefix filter list create, MCR IPSec add-on add, and the apply order paths.
- Fixed inconsistent outliers: LAG buy and IX buy previously had no retry; IX update/delete now use the standard mutation retry like their port/MCR peers.
- Reads and idempotent updates (update/delete/modify) keep their existing helpers.
- Tests: unit tests verified load-bearing via mutation check, `--no-retry` path covered, plus a package-level test wiring a mock IX buy through the new path.

Closes ESD-1500.

### DIFF
--- a/internal/utils/retry.go
+++ b/internal/utils/retry.go
@@ -34,6 +34,10 @@ type RetryOpts struct {
 	// RetryNetworkErrors enables retrying on ambiguous network failures
 	// (connection reset, EOF, timeout). Only set this for idempotent operations.
 	RetryNetworkErrors bool
+	// OrderSafe restricts retries to errors that prove the request never reached
+	// processing (HTTP 429, connection-refused-before-send). Set this for
+	// non-idempotent buy/order calls so an ambiguous failure cannot double-submit.
+	OrderSafe bool
 }
 
 // DefaultRetryOpts returns sensible defaults for API retry behaviour.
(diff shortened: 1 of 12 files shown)


internal/utils/retry.go
```diff
@@ -75,6 +79,21 @@ func WithIdempotentRetry(ctx context.Context, fn func(ctx context.Context) error
 	return RetryWithBackoff(ctx, opts, fn)
 }
 
+// WithOrderRetry wraps fn with a retry policy safe for non-idempotent buy/order
+// operations. It retries ONLY when the request provably never reached
+// processing: HTTP 429, or a connection-refused error raised before the request
+// was sent. It never retries on 5xx or other ambiguous-outcome errors, which
+// could double-submit an order and cause duplicate billing.
+// If --no-retry was set globally, fn is called exactly once.
+func WithOrderRetry(ctx context.Context, fn func(ctx context.Context) error) error {
```
ianb-mp, line 88: Naming this `WithOrderOnceRetry` may convey the intent more clearly (and link it semantically to the `OrderOnce` bool).

internal/utils/retry.go
```diff
@@ -193,6 +217,19 @@ func isRetryable(err error, retryNetworkErrors bool) bool {
 	return false
 }
 
+// isOrderRetryable returns true only for errors that prove a buy/order request
+// never reached processing: HTTP 429, or connection-refused raised during the
+// TCP handshake (the request bytes never left the client). Ambiguous outcomes
+// (5xx, connection reset, timeout, EOF) are never retried because the order may
+// already have been accepted upstream.
+func isOrderRetryable(err error) bool {
```
ianb-mp, line 225: Renaming this to `isRetryableOrderOnce` may provide a clearer link to the `OrderOnce` bool.

internal/utils/retry.go
```diff
@@ -34,6 +34,10 @@ type RetryOpts struct {
 	// RetryNetworkErrors enables retrying on ambiguous network failures
 	// (connection reset, EOF, timeout). Only set this for idempotent operations.
 	RetryNetworkErrors bool
+	// OrderSafe restricts retries to errors that prove the request never reached
+	// processing (HTTP 429, connection-refused-before-send). Set this for
+	// non-idempotent buy/order calls so an ambiguous failure cannot double-submit.
+	OrderSafe bool
```
ianb-mp, line 40: Naming this something like `OrderOnce` might convey clearer intent.

ianb-mp (approved):
lgtm, just some queries on naming of things.

penzeliz-megaport (approved):

PR #427, opened 2026-06-16
fix: print command errors to stderr and keep stdout data-only

Phil-Browne:
Two output-contract bugs (ESD-1499): failing commands could exit non-zero while printing nothing, and status messages leaked onto stdout in non-json formats, polluting redirected data files.

- All RunE wrapper error handling now funnels through `finishWithError`: it prints the error to stderr for non-json formats and emits a single JSON envelope for json. An error-emitted latch stops it double-printing when an action already showed the error.
- `Print{Error,Warning,Info,Success}` always write to stderr now, in every format, so stdout carries only data. `PrintError` is a no-op in json (the wrapper owns the envelope).
- Each wrapper syncs the output package format from the resolved flag, so a stale json format in the long-lived WASM process can't swallow a non-json command's error.
- Tests across the command packages updated to assert messages on stderr.

Verified both repros: `mcr update-prefix-filter-list someuid notanumber` now prints to stderr (and a single JSON envelope under `--output json`); `ports list --output JSON` prints its usage error instead of vanishing. `go test ./...`, `go build ./...`, `golangci-lint run`, and the WASM build all pass.


penzeliz-megaport (dismissed):
LGTM. Approving with the backward-incompatibility accepted deliberately: this makes stdout data-only and routes all status/info/warning/error output to stderr across every command. The `finishWithError` double-print guard is well-tested (count==1 when an action self-prints; silent errors still surface), and the `errorEmitted` global latch is reset per-invocation via `ResetState`/`ResetErrorEmitted`, which is safe for the one-command-per-process CLI and serial WASM invocations.

Two things worth a follow-up (non-blocking):
1. Confirm no first-party consumer (the WASM web UI, e2e scripts) scrapes success/info lines off stdout — those now go to stderr.
2. `CaptureOutput` changed semantics (now stdout+stderr, was stdout-only). The known call sites are updated and `CaptureStdout` added for data-only cases, but the silent semantic shift of an existing helper is a footgun for future tests — worth a shared `testutil` helper to kill the per-package `captureStderr` duplication.

Approving.

penzeliz-megaport (dismissed):
Re-confirmed at the current head: the only change since my prior approval is a main-merge (conflict resolution in `servicekeys_actions_test.go`/`vxc_actions_test.go` just added main's new `*_NilResponse` tests alongside the existing stderr-assertion tests). All of the subject output-package files are byte-identical. Approval stands, carrying the same two non-blocking notes: confirm no first-party consumer scrapes status lines off stdout (they're now on stderr), and the `CaptureOutput` semantic shift + per-package `captureStderr` duplication are worth a shared `testutil` helper later.

Heads-up: branch currently shows as conflicting (main advanced after #449 merged) — needs a fresh main-merge before it can land.

penzeliz-megaport (approved):

PR #428, opened 2026-06-18
ci: add GitHub Action to build and publish WASM to S3 (ESD-1506)

Phil-Browne:
Adds a GitHub Action that builds the browser/WASM CLI static site and publishes it to a CloudFront-fronted S3 bucket, replacing the manual upload step.

**What it does**
- Manual-only (`workflow_dispatch`): a production publish is always a deliberate action, never an automatic tag push.
- Builds the static site with `make web-static`, then pre-compresses the wasm to brotli (CloudFront auto-compression caps at 10 MB; the raw wasm is larger).
- Publishes stable filenames (`megaport.wasm` served with `Content-Encoding: br`, `wasm_exec.js`, and the rest of the site) under a versioned prefix plus a short-cached `latest/` alias.
- Authenticates via AWS OIDC (no long-lived keys); pins all actions to commit SHAs.

Filenames are kept stable (not content-hashed) because the portal references them by static config URL; the version/latest prefix is the cache-buster instead.

Required repo config (secrets and variables) is provisioned out of band and validated by the workflow's config-check step, which fails fast with a clear message if anything is missing.

ESD-1506

### DIFF
--- a/.github/workflows/wasm-publish.yaml
+++ b/.github/workflows/wasm-publish.yaml
@@ -8,8 +8,10 @@ name: Publish WASM to S3
 #   megaport/megaport-cli to github_repo_to_s3_prod_deploy_role_mappings granting the
 #   media.megaport.com bucket. That provisions this repo's standard
 #   AWS_S3_PROD_DEPLOY_ROLE. Then set this repo config:
-#     secrets.AWS_S3_PROD_DEPLOY_ROLE  role ARN to assume (a secret only to keep the AWS
-#                                      account id out of the plaintext variables list)
+#     secrets.AWS_S3_PROD_DEPLOY_ROLE  role ARN to assume (a secret so the embedded AWS
+#                                      account id stays masked in this public repo's
+#                                      world-readable Actions logs; OIDC trust is the real
+#                                      guard, this is just defense in depth)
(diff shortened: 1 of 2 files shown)


CpuID (dismissed):

penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. Security hygiene is notably good: `workflow_dispatch` only (no `pull_request_target`, no auto tag-push), least-privilege permissions (`contents: read`, `id-token: write`), AWS auth via OIDC role-assumption (no long-lived keys), all three actions SHA-pinned, the version label that becomes an S3 key / CloudFront path is validated (rejects `..`, enforces `[A-Za-z0-9._-]`), and untrusted inputs are passed via `env` rather than inlined into `run:`. Approving.

Two operational items to track as follow-ups (acknowledged in the PR's SAFETY comment, not blocking):
1. No approval gate on a production deploy — the shared OIDC role trusts the whole repo, so any write-access user can dispatch. The suggested `production` environment with required reviewers is the right next step.
2. `aws s3 sync --delete` against a shared bucket prefix — keep `WASM_S3_PREFIX` tightly scoped so a misconfigured var can't wipe sibling content.

Approving on the basis that prefix scoping is confirmed correct and the no-gate posture is accepted for now.

PR #430, opened 2026-06-18
fix(locations): drop blank market codes from list-market-codes

Phil-Browne:
`locations list-market-codes` rendered an `<empty>` row (and counted it in the total). The network regions API returns some countries with an empty market prefix, and the SDK passes those straight through.

This filters out blank/whitespace market codes in the `ListMarketCodes` action, before the count message and the output, so both stay consistent.

- Drop blank/whitespace codes returned by the SDK
- Add a `TestListMarketCodes` case covering empty and whitespace-only entries

ESD-1509

### DIFF
--- a/internal/commands/locations/locations_actions.go
+++ b/internal/commands/locations/locations_actions.go
@@ -3,6 +3,7 @@ package locations
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):
LGTM

PR #434, opened 2026-06-18
ci: add WASM build guard and browser fetch smoke harness (ESD-1286)

Phil-Browne:
CI never compiled the `js/wasm` target, so a change that broke the browser build (or a newly registered module that doesn't compile under WASM) could ship green. This adds the missing safety net plus the contributor docs for enabling a module in WASM.

**Changes**
- New `wasm` workflow with two jobs, kept separate from the native matrix so a WASM-only failure is obvious in the checks list:
  - `build-guard` compiles the `js/wasm` target on every PR and fails on breakage.
  - `smoke` runs `locations list` end-to-end through the WASM fetch transport under Node 24 against staging, proving a command actually round-trips (not just links). Uses the unauthenticated locations endpoint, so no credentials are needed.
- `make wasm-build-guard` and `make wasm-smoke` targets, plus `scripts/wasm-smoke.mjs` (the Node harness) for local runs.
- Per-module "add to WASM" recipe in `WASM_README.md` (register, build-guard, optional `_actions_wasm.go` override, smoke). Clarifies when an override is actually needed vs not.
- Fixed the stale module list in `WASM_README.md` to match `modules_wasm.go` (nat-gateway, product, status, topology are registered).
- Removed `internal/commands/locations/locations_actions_wasm.go`: the override was redundant since `config.NewUnauthenticatedClient()` already uses the fetch transport in WASM. The native code path works without it.

The smoke harness loads the same `WasmHTTPTransport` the browser uses, so it exercises the real fetch path. It round-trips ~610 staging locations in CI today.

Ticket: ESD-1286


penzeliz-megaport (approved):
LGTM on the CI/security axis. `permissions: contents: read`, all actions SHA-pinned, `pull_request` (not `pull_request_target`), and the smoke job hits only the unauthenticated staging `locations` endpoint so no credentials are wired in (and the docs explicitly warn against adding any). The `build-guard` + Node WASM `smoke` split is sensible. The deletion of `locations_actions_wasm.go` and the Node harness correctness are functional rather than security concerns; nothing in the workflow surface blocks. Minor doc nit (non-blocking): README says Node 22+ while the workflow uses Node 24. Approving.

PR #436, opened 2026-06-18
ci: add govulncheck workflow

Phil-Browne:
Adds a `govulncheck` CI job so we get alerted to known vulnerabilities in the Go stdlib and our dependencies, on every push to main and PR.

It mirrors the existing lint/test workflows: pinned action SHAs, `go-version-file`, read-only permissions, and a pinned govulncheck version to avoid CI drift.

Ran it locally against the current tree under the CI toolchain (go1.26.4): **no vulnerabilities found**, so this lands green. govulncheck is call-graph aware, so it only flags vulns our code actually reaches.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -1,0 +1,24 @@
+name: govulncheck
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  govulncheck:
+    name: govulncheck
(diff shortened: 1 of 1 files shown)


penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. Minimal `contents: read` permissions, both actions SHA-pinned, `govulncheck` pinned to v1.4.0, and it triggers on `pull_request` (not `pull_request_target`) so forks run sandboxed with a read-only token. No secrets, nothing to escalate. Approving.

PR #438, opened 2026-06-19
fix(output): stop spinner animating into non-TTY logs

Phil-Browne:
When output is not a terminal (CI logs, pipes, file redirects), the spinner kept animating: it wrote a `\r\033[K<frame> <msg>` line every ~100ms, and since the carriage-return/clear-line escapes don't collapse in a non-TTY sink, every frame became a fresh line (the hundreds of `Validating Port order...` lines in the provisioning integration logs). The `!IsTerminal()` check only chose stderr vs stdout; it never stopped the animation.

Now the spinner detects a non-interactive sink (non-table machine format, or not a TTY) and prints its status message once to stderr instead of starting the ticking goroutine. `Stop()` only emits the bare clear-line escape on the interactive path, so it no longer leaves junk in logs.

- Add `(*Spinner).nonInteractive()` and use it in `runLoop`, `Stop`, `StopWithSuccess`.
- Non-interactive `runLoop` prints the prefix once to stderr and skips the goroutine.
- Interactive TTY behavior (in-place animated spinner) is unchanged. WASM path untouched.
- New `TestSpinnerNonTTYDoesNotAnimate` (forces `IsTerminal()` false) asserts the message appears once with no `\r` / `\033[K` spam.

Verified: `go test ./...`, `go build ./...`, native + WASM build, and `golangci-lint run` all clean. Sanity-checked by piping a spinner command to `cat`: one static status line, zero CR / clear-line bytes on stdout.

Fixes ESD-1515.

### DIFF
--- a/internal/base/output/messages.go
+++ b/internal/base/output/messages.go
@@ -239,6 +239,14 @@ func (s *Spinner) renderFrame(i int) string {
 	return color.CyanString(frame)
 }
 
+// nonInteractive reports whether the spinner's sink is non-interactive: a
+// machine-readable output format, or output not attached to a TTY. In those
+// sinks the carriage-return/clear-line escapes don't collapse anything, so the
+// spinner emits a single plain status line rather than animating frame-by-frame.
+func (s *Spinner) nonInteractive() bool {
+	return (s.outputFormat != "" && s.outputFormat != "table") || !IsTerminal()
+}
+
 // runLoop is the shared spinner goroutine logic. If startTime is non-nil,
 // an elapsed duration is appended to the message on each tick.
 func (s *Spinner) runLoop(prefix string, startTime *time.Time) {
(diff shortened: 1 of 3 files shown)


penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. The `nonInteractive()` predicate is a clean extraction of the existing inline condition, applied consistently across `runLoop`/`Stop`/`StopWithSuccess`, and the test asserts the message prints exactly once with no `\r`/`\x1b[K` leakage. Confirmed the non-interactive path is deadlock-free: `stop` is `make(chan bool, 1)` (buffered cap 1), so the unconditional `s.stop <- true` in `Stop()` fills the buffer slot rather than blocking when the animation goroutine was never started, and the `stopped` guard ensures it fires at most once. Approving.

PR #440, opened 2026-06-19
fix(apply): wire MCR tunnel_count to IPsec add-on order

Phil-Browne:
The `apply` command silently dropped an MCR's IPsec tunnel count: `MCRConfig` had no `tunnel_count` field, and the config decoder ignored unknown keys, so the value just vanished and the MCR came up with no IPsec add-on (#439).

- Add `tunnel_count` to the MCR config (valid values: 10, 20, 30; 0 or absent means no IPsec add-on)
- Wire it through to the MCR order in both the provisioning and dry-run paths, with the add-on type set explicitly
- Validate the count client-side on both paths, so a bad value fails fast instead of reaching the API
- Make the config decoder strict: unknown keys (e.g. the camelCase `tunnelCount` from the original report), trailing JSON data, and stray extra YAML documents now fail loudly with the offending field named, instead of being silently dropped

SDK support already shipped in megaportgo (v1.13.1), so this is CLI-only.

Note: the apply schema uses snake_case, so the key is `tunnel_count`, not `tunnelCount`. A fuller apply-schema doc/example is worth a follow-up; strict decoding already turns a mistyped key into a clear, self-correcting error.

Closes #439


penzeliz-megaport (dismissed):
LGTM. The `tunnel_count` wiring into both the provisioning and dry-run/validate paths via `mcrAddOns()` mirrors `mcr buy` and is well-tested. Approving with the strict-decoding change accepted deliberately: `DisallowUnknownFields`/`KnownFields(true)` means any previously-tolerated unknown/misspelled key in an existing apply config now hard-fails. Good that `TestParseConfigFile_AllowsFreeformMapKeys` confirms strictness applies only to struct fields, not to `resource_tags`/`vendor_config` map values, and benign trailing content is allowed. Confirmed `go.mod` pins megaportgo v1.13.1, which carries the MCR IPsec add-on types.

Follow-up (non-blocking but worth landing soon): since strict decoding now hard-errors on the camelCase `tunnelCount` form, the snake_case `tunnel_count` docs/example should ship alongside so users don't hit the error blind. Approving.

penzeliz-megaport (approved):
Re-confirmed at the current head: the only change since my prior approval is a main-merge (conflict resolution confined to `apply_mock.go` absorbing the new `GetXxxReturnNil` mock hooks); the `tunnel_count` wiring and strict-decode logic are byte-identical. Approval stands.

Still-open follow-up (non-blocking, accepted at review): the snake_case `tunnel_count` docs/example still aren't present anywhere under `docs/`/`*.md`. Since the same change hardens the decoder to reject the camelCase `tunnelCount` people will naturally try (issue #439 was exactly that), please land a one-line apply-config example soon after merge so users don't hit the strict reject blind.

PR #441, opened 2026-06-19
feat(wasm): register users, managed-account, billing-market modules

Phil-Browne:
Registers the `users`, `managed-account`, and `billing-market` command modules in the WASM browser build so it reaches parity with the native CLI (ESD-1287). All three already use SDK-direct calls with no filesystem or `os/exec` dependencies, so this is a register-and-go change with no `_wasm.go` override needed.

Changes:
- Add the three modules to `registerModules()` in `cmd/megaport/modules_wasm.go`.
- Add `TestAccountPartnerAdminModulesRegistered`, a help-only smoke test asserting each module (plus one read-only subcommand each) is wired into the WASM command tree.
- Add a `wasm` CI workflow that builds the browser artifact and runs the WASM unit tests for `./cmd/megaport` under the node runtime.

CI now compiles the js/wasm build and runs the `./cmd/megaport` WASM tests (including the new smoke test). The other WASM test packages need a real browser host (fetch, JS bridge globals, stdin pipes), so they stay out of CI for now. The in-browser checks against Staging called out in the ticket (each command working, a non-partner key on managed-account behaving sensibly rather than crashing, and mutating commands confirmed with no prod side effects) remain manual.

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -1,0 +1,29 @@
+name: wasm
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  wasm:
+    name: WASM Build & Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5.0.1
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
+        with:
+          go-version-file: 'go.mod'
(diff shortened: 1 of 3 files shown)


penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. Registration of the `users`, `managed_account`, and `billing_market` modules mirrors the native registration exactly, and all three are SDK-direct (no `os/exec`/filesystem imports) so they're WASM-safe. Help-only smoke test asserts on the `Usage:` block. Note for the merge queue: this and #422 both edit `cmd/megaport/modules_wasm.go`, so whichever lands second needs a trivial rebase. Approving.

PR #444, opened 2026-06-22
fix: classify ValidationError as usage exit code (ESD-1523)

Phil-Browne:
`classifyError` picked the exit code by matching the lowercase substring `"invalid"` and also requiring `"ID"`. A raw `ValidationError` formats its message as `"Invalid %s: ..."` (capital I), so it never matched and the process exited with code 1 (General) instead of 2 (Usage). Scripts couldn't tell a usage/validation failure from a generic one.

The fix maps the error by type instead of by message text:

- Add an `errors.As(err, &validationErr)` check in `classifyError` that returns `exitcodes.Usage` for any `ValidationError`, placed right after the existing `CLIError` passthrough (which still wins).
- Substring heuristics and the `ValidationError` message format are left untouched.
- Add `TestClassifyError_ValidationError` covering the bare and `%w`-wrapped cases.

`go build ./...`, `go test ./...`, and `golangci-lint run` are clean.

Jira: ESD-1523

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/megaport/megaport-cli/internal/base/exitcodes"
 	"github.com/megaport/megaport-cli/internal/base/output"
+	"github.com/megaport/megaport-cli/internal/validation"
 	megaport "github.com/megaport/megaportgo"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. The `errors.As(&validationErr)` type check is the right fix for the capital-I `Invalid %s` strings that the lowercase `"invalid"` substring heuristic never matched, and placing it after the `CLIError` passthrough preserves explicit exit codes. Covers both bare and `%w`-wrapped cases plus an end-to-end `WrapRunE` test. Pre-existing substring heuristics left in place is fine (out of scope). Approving.

PR #445, opened 2026-06-22
fix: apply --timeout bounds each resource, not the whole run (ESD-1522)

Phil-Browne:
`apply` used to create one context (default 10m, or `--timeout`) and share it across every resource's provisioning wait. On a large plan the run could time out partway through even though each resource was provisioning fine, because they all drew from one budget. The per-resource fallback in `waitForProvision` only kicked in when the context had no deadline, which never happened.

This makes `--timeout` a per-resource budget instead of a whole-run one.

- Run-wide context is now deadline-free; each `waitForProvision` gets its own `context.WithTimeout(timeout)`.
- Individual API calls stay bounded by the SDK HTTP client timeout (30s), so nothing runs unbounded.
- Rollback keeps its own independent context and reverse-order unwind, unchanged.
- Kept the `hasDeadline` fallback in `WaitForProvision`: it still caps any deadline-free caller at `DefaultProvisionTimeout`.
- Documented the per-resource semantics in the `apply` long description.

New test `TestApplyConfig_PerResourceTimeout` provisions two resources that each consume most of the budget; their combined wait exceeds a single shared budget but each stays within its own, so apply completes only under the new behavior.

Jira: ESD-1522


penzeliz-megaport (dismissed):
LGTM. Approving with the semantics change accepted deliberately: `--timeout` now bounds each resource's provisioning wait rather than the whole run, so a multi-resource apply can take up to N× the value. The run-wide context is now deadline-free (`WithCancel(Background())`) with each resource deriving its own `WithTimeout`, and individual API calls remain bounded by the SDK's HTTP client timeout, so nothing runs truly unbounded. `TestApplyConfig_PerResourceTimeout` genuinely proves the new behaviour (two ~180ms resources pass under a 300ms per-resource budget that a shared budget would fail). Rollback correctly reuses `provisionTimeout`.

Non-blocking: the test is wall-clock based (180ms/300ms with a 2ms poll) and not guarded by `testing.Short()`, so it could flake on a heavily loaded CI runner. The doc update covers the behaviour change. Approving.

penzeliz-megaport (dismissed):
Re-confirmed at the current head: the only change since my prior approval is a main-merge (conflict resolution confined to `apply_mock.go`); the per-resource-timeout logic in `apply_actions.go` is byte-identical. Approval stands, carrying the same non-blocking note: `TestApplyConfig_PerResourceTimeout` is wall-clock based (180ms/300ms, no `testing.Short()` guard) so it could flake on a loaded CI runner.

Heads-up: branch currently shows as conflicting (main advanced again after #449 merged) — needs a fresh main-merge before it can land.

penzeliz-megaport (approved):

PR #446, opened 2026-06-22
fix(json): tighten JSON input value handling to match flags (ESD-1521)

Phil-Browne:
Tightens the `--json` / `--json-file` input path so it validates values as strictly as the flag and update paths.

- VXC buy now rejects fractional `term`/`rateLimit` (e.g. `10.9`) instead of silently flooring to `10`; the same `rateLimit` whole-number check is added to VXC update.
- VXC buy, MVE buy, and MVE update now return a clear `X must be a <type>` error when a JSON field is present but the wrong type, instead of silently dropping it. Absent optional keys stay optional.
- The tag JSON parser now rejects an empty key (`{"":"x"}`), matching interactive entry which treats an empty key as "stop".

Mechanism is a shared `internal/utils/json_map.go` helper set (`JSONString/JSONNumber/JSONBool/JSONObject/JSONArray`) returning `(value, present, err)`.

Two deliberate boundaries, both for compatibility:
- VXC update keeps only the fractional-number check, not the wrong-type errors, because of its dual-key fallbacks (`name`/`vxcName`, `aEndConfiguration`/`aEndVlan`).
- The embedded `resourceTags` map in buy commands is left as-is. Empty-key rejection there would need to land uniformly across all buy commands (which decode tags differently), so it is tracked as a follow-up.

### DIFF
--- a/internal/commands/mcr/mcr_inputs.go
+++ b/internal/commands/mcr/mcr_inputs.go
@@ -26,6 +26,10 @@ func processJSONMCRInput(jsonStr, jsonFile string) (*megaport.BuyMCRRequest, err
 		return nil, fmt.Errorf("failed to parse JSON: %w", err)
 	}
 
+	if err := utils.RejectEmptyTagKeys(req.ResourceTags); err != nil {
+		return nil, err
+	}
+
 	// BuyMCRRequest.AddOns is []MCRAddOn (interface) and cannot be directly
 	// unmarshaled by the standard library. Read tunnelCount separately using a
 	// pointer so we can distinguish "key absent" from an explicit value.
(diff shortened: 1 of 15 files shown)


penzeliz-megaport (approved):
LGTM. Approving with the input-validation tightening accepted as intended: wrong-typed-but-previously-ignored JSON fields, and fractional `rateLimit`/`term` values that used to floor silently, now error. The shared `utils.JSON*` helpers returning `(value, present, err)` are clean and the round-trip tests are thorough. Confirmed `"math"` is imported in `vxc_inputs_update.go` (it compiles).

Two non-blocking notes carried over from review:
1. Self-acknowledged asymmetry: VXC update gets only the fractional check and still uses a raw `rawData["rateLimit"].(float64)` cast rather than `utils.JSONNumber`, so a wrong-typed `rateLimit` on update is still silently dropped while on buy it errors. Worth a follow-up for consistency.
2. `resourceTags` embedded in buy commands still skips empty-key rejection (only the standalone parse paths got it) — partial, tracked as follow-up.

Approving.

PR #447, opened 2026-06-22
fix: guard nil SDK responses and honor vNIC index 0 on VXC (ESD-1524, ESD-1525)

Phil-Browne:
Two small robustness fixes, swept across every path that had the same shape.

**Nil response guards.** A number of paths dereferenced the SDK response without checking it was non-nil, so a `(nil, nil)` return would panic. They now return a clear "empty response from API" error instead, matching the existing ports/MCR/MVE guards. Covered:

- VXC buy, IX buy
- All four service-key paths (create, get, update, list)
- User create and managed-account create
- The `apply` provisioning polls (port, MCR, MVE, VXC), which read status off the polled resource

**vNIC index 0.** Building a VXC with an MVE endpoint at vNIC index 0 silently dropped that endpoint, because the code gated on a non-zero index. Fixed on both input modes that had it: the flag path (gates on whether the flag was set) and the interactive prompt (gates on whether an index was entered). The JSON path already keyed off presence and is unchanged.

Notes:
- Out of scope by design: `ValidateVNICIndex` and the JSON parse path are untouched.
- Sibling update/get paths that feed nil into helpers already guarding nil were left as-is, so they don't panic.

Every guarded path and the vNIC fix have tests; each was confirmed to fail (panic, or drop the endpoint) before the fix. `go build`, `go test ./...`, and `golangci-lint` are clean.

### DIFF
--- a/internal/commands/apply/apply_actions.go
+++ b/internal/commands/apply/apply_actions.go
@@ -194,6 +194,9 @@ func ApplyConfig(cmd *cobra.Command, _ []string, noColor bool, outputFormat stri
 			if e != nil {
 				return "", e
 			}
+			if port == nil {
+				return "", fmt.Errorf("empty response from API")
+			}
 			return port.ProvisioningStatus, nil
 		}); err != nil {
 			createSpinner.Stop()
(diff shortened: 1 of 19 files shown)


penzeliz-megaport (approved):
LGTM. Two solid robustness fixes. The nil-response guards are purely additive (panic → clean `empty response from API` error) across all the affected paths. The vNIC-0 fix is correct by construction: gating on `cmd.Flags().Changed("a-end-vnic-index")` (flag path) / non-empty input (prompt path) is the right "was a value supplied" semantics regardless of whether 0 is a valid index — the old `index > 0` gate silently dropped a legitimately-supplied index 0, which this fixes, and it now matches the already-presence-keyed JSON path. Well-tested on both halves.

Merge-order note: this and #449 both edit the `apply_actions.go` provision polls and will textually conflict. Recommend merging this first; #449 relocates these callbacks into `utils.WaitForProvision` and carries equivalent nil checks, so the guard intent survives the rebase. Approving.

PR #448, opened 2026-06-22
fix: reject explicit non-positive --timeout and clarify help (ESD-1526)

Phil-Browne:
Fixes ESD-1526.

The global `--timeout` help claimed `0 uses the internal default of 90s`, but 90s is only the default for read-only commands. Mutation commands default to 15m, `topology` to 120s, one `vxc` path to 5m, and `apply` to 10m (and treats `--timeout` as a per-resource budget since ESD-1522). The single "90s" claim was wrong for most commands, and the "request timeout" framing was wrong for `apply`.

On review, `--timeout 0` was also treated as a silent "use the default" sentinel, which hides user mistakes. It is now a usage error: omit the flag to get a command's default, or pass a positive duration.

- Reword the help: `Timeout for the operation (...); must be positive. Omit to use each command's built-in default (see the command's own help)`.
- Drop the misleading "request timeout" / fixed-90s wording.
- Reject an explicit non-positive `--timeout` via a shared `utils.ValidateTimeoutFlag` helper (uses `FlagSet.Changed`), wired into `PersistentPreRunE` for native and WASM, alongside the existing `--max-retries` check.
- Add unit tests for the helper and an end-to-end test that `version --timeout 0` exits with a usage error.

### DIFF
--- a/cmd/megaport/common_wasm.go
+++ b/cmd/megaport/common_wasm.go
@@ -81,7 +81,7 @@ func InitializeCommon() {
 	rootCmd.PersistentFlags().StringVar(&utils.Env, "env", "", "Environment to use (prod, dev, or staging)")
 	rootCmd.PersistentFlags().BoolVarP(&quiet, "quiet", "q", false, "Suppress informational output, only show errors and data")
 	rootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Show additional debug information")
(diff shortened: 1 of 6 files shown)


mp1-tomclancy (dismissed):
Looks good.

However, I wonder if the semantics of `--timeout` would be clearer with the following logic:
- if not supplied, use default values
- if supplied, validate that `timeout >0`

[FlagSet.Changed](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed) might be an option here.

I would argue that the following should be invalid, rather than default to `5m`:

```
megaport-cli vxc update --timeout 0 ...
```

mp1-tomclancy (approved):

PR #449, opened 2026-06-22
fix(buy): submit order under retry, poll provisioning separately (ESD-1520)

Phil-Browne:
Every standalone `buy` command (ports, vxc, mcr, mve, ix) set `WaitForProvision=true` on the order request *and* wrapped the whole SDK buy call in `WithOrderOnceRetry`. That SDK call both submits the order and polls until provisioned, so a 429 raised during the poll phase retried the entire call and re-submitted the order, risking duplicate billing.

This splits the two: submit the order under the retry, poll for provisioning afterwards outside it.

- Set `WaitForProvision=false` on all five buy paths; only the order submission is wrapped in `WithOrderOnceRetry`.
- Poll for provisioning afterwards via a shared `utils.WaitForProvision` helper, factored out of the `apply` path (no duplicate poll loop).
- Default provisioning wait, the 10-minute cap, and `--no-wait` are unchanged from the user's perspective.
- Minor behavior change: polling now runs CLI-side at 10s intervals (matching `apply`) instead of the SDK's old 30s; the 10-minute cap is unchanged.
- Added a regression test per buy path: a retryable 429 during the poll surfaces as an error and never re-submits the order.


penzeliz-megaport (dismissed):
LGTM, and the core claim checks out. I verified against megaportgo v1.13.1 source: with `WaitForProvision=false` the SDK returns immediately after `ExecuteOrder` (no internal poll loop), so wrapping only the submit call in `WithOrderOnceRetry` means a 429 during the now-CLI-side provisioning poll surfaces as an error and can no longer re-submit the order. That's the duplicate-billing fix and it's sound. Per-path 429-during-poll regression tests + the `utils.WaitForProvision` helper unit tests cover ready/error/terminal/poll/timeout — good coverage.

Merge-order note: depends on #447 (both edit the `apply_actions.go` polls). Land #447 first, then rebase this; the relocated `utils.WaitForProvision` callbacks already carry equivalent nil guards so #447's intent survives. Approving.

penzeliz-megaport (approved):

PR #450, opened 2026-06-22
fix(tags): reject empty tag keys in buy/create JSON input (ESD-1527)

Phil-Browne:
Empty tag keys (`{"": "x"}`) were already rejected by `update-tags` (ESD-1521) but the buy/create commands' embedded `resourceTags` still accepted them. This makes the rejection uniform.

- Export the existing `rejectEmptyTagKeys` helper as `utils.RejectEmptyTagKeys` (same `tag key must not be empty` message, no duplicated logic).
- Call it on the populated tag map (before the `[]ResourceTag` conversion) in: `buy vxc`, `buy mcr`, `buy port`, and both `create` and `update` nat-gateway JSON paths.
- Unit test per path, including `--json-file` coverage and a valid-tags-survive case.

Stacked on #446 (ESD-1521); base is `worktree-ESD-1521-json-input-hardening`, not `main`.

### DIFF
--- a/internal/commands/mcr/mcr_inputs.go
+++ b/internal/commands/mcr/mcr_inputs.go
@@ -26,6 +26,10 @@ func processJSONMCRInput(jsonStr, jsonFile string) (*megaport.BuyMCRRequest, err
 		return nil, fmt.Errorf("failed to parse JSON: %w", err)
 	}
 
+	if err := utils.RejectEmptyTagKeys(req.ResourceTags); err != nil {
+		return nil, err
+	}
+
 	// BuyMCRRequest.AddOns is []MCRAddOn (interface) and cannot be directly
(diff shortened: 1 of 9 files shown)


ianb-mp (approved):

penzeliz-megaport (approved):
LGTM. Exporting `rejectEmptyTagKeys` as `utils.RejectEmptyTagKeys` (no logic change, message preserved, both internal callers updated) and calling it on the populated tag map in `buy vxc/mcr/port` and nat-gateway create+update JSON paths is the right way to close the buy/create gap left by ESD-1521. Called at the correct point (post-populate, pre-`[]ResourceTag` conversion), with a unit test per path including `--json-file` coverage and a valid-tags-survive case.

Merge-order note: this is stacked on the #446 branch (now approved), not main — it must merge after #446, and its diff against main will settle once it rebases. Approving.

PR #451, opened 2026-06-22
test: add VLAN-scoped and update-matrix service key integration tests (ESD-1530)

Phil-Browne:
Broadens service-key lifecycle integration coverage for ESD-1530.

- `TestIntegration_ServiceKeyVLANScopedLifecycle`: buys a port, creates a single-use VLAN-scoped key, asserts VLAN + single-use + product round-trip via the SDK.
- `TestIntegration_ServiceKeyUpdateMatrix`: creates a key, confirms create round-trips its fields, normalizes it to a known active+single-use baseline, then single-field updates `active` then `single-use`, asserting via the SDK that the sibling it did not name survives. This is the ESD-1417 guard against unset bools resetting sibling fields. `max-speed` is checked as a preserved sibling on every update.

Both are tagged `integration && provisioning`, so they run only in the manual provisioning job, not the read-only nightly. `go build` and untagged `go test` pass; vet and lint are clean under the tags. Both tests pass against staging.

Two SDK realities worth calling out, since they diverge from the ticket's literal wording:

- The update API exposes only `active`/`single-use` (no `max-speed` or `description` fields). So those bools are toggled directly, while `max-speed` and `description` are checked for survival rather than updated. The API preserves `max-speed` across an update, but **clears `description`**: the update request carries no description field, so any update (even toggling `active`) wipes a key's description. The test pins this as a documented limitation (`assert.Empty` on description after the first update). Round-tripping description on update would need an SDK change and is a sensible follow-up.
- Service keys have no delete API, so cleanup tears down only the backing port; the key is left attached to the deleted port. This matches the existing `ServiceKeyLifecycle` test. Over many runs these keys accumulate on staging by design.

Also adds the `RequireStagingForProvisioning` guard to the older `ServiceKeyLifecycle` test, which was missing it. All three provisioning tests now refuse to run outside staging, so the suite can never buy a real port against production.

### DIFF
--- a/internal/commands/servicekeys/servicekeys_lifecycle_integration_test.go
+++ b/internal/commands/servicekeys/servicekeys_lifecycle_integration_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


penzeliz-megaport (approved):
LGTM. The update-matrix test normalises to a known active+single-use baseline before asserting, so the ESD-1417 sibling-reset regression can't pass trivially, then checks the unspecified sibling survives in both update directions. `assert.Empty(description)` after the first update pins the real documented limitation (updates clear description) rather than masking it. Adding the missing `RequireStagingForProvisioning` guard to the older `ServiceKeyLifecycle` test closes a genuine gap. Service keys having no delete API (so keys accumulate on staging, cleanup only tears down the backing port) matches existing behaviour. Approving.

PR #452, opened 2026-06-22
test: add LAG port update lifecycle integration test (ESD-1528)

Phil-Browne:
Adds `TestIntegration_LAGPortUpdateLifecycle`. It buys a LAG port and exercises the update path the existing name-only LAG test skips: flip marketplace visibility, set a cost centre, each verified through an SDK `GetPort` read. Delete and decommissioning ride on the existing `t.Cleanup` poll helper.

LAG member count and port speed turn out to be immutable via the update action (no field on `ModifyPortRequest`, no update flag), so the test scopes to the mutable set and records the immutable attributes in a comment.

- New `integration && provisioning` test in `ports_integration_test.go`
- Retag the file from `integration` to `integration && provisioning` so the provisioning lifecycle tests are kept out of the nightly read-only build by the tag itself, not just the `-run '...ReadOnly$'` filter (matches the servicekeys/users convention)
- Doc update: describe the ports tag and flag VXC/MCR/MVE/IX (IX especially) as a follow-up to align the same way

Validation: `go build` and `go test` (no tag) pass; vet/lint clean under `integration provisioning`; tag-discovery confirms the new test runs only in the provisioning build and is excluded from nightly read-only. The test provisions a real staging LAG port, so it runs in the manual provisioning job (not nightly).

### DIFF
--- a/docs/INTEGRATION_TESTING.md
+++ b/docs/INTEGRATION_TESTING.md
@@ -61,7 +61,7 @@ A lifecycle test that lives in a package the nightly read-only job also runs car
 package servicekeys
 ```
 
-The nightly read-only job builds only `-tags integration`, so this tag keeps the mutating test out of it. It is needed only for packages that the read-only job runs — currently `servicekeys` and `users`, which have read-only `list`/`get` tests nightly but whose lifecycle tests provision resources (the service key test buys a port to use as its product). Lifecycle tests in packages the read-only job does not run (ports, VXC, MCR, MVE, IX) are excluded from nightly by package selection alone, so they stay `//go:build integration` only.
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. Good test discipline: asserts the buy-time baseline (visibility=false, speed 10000) before each update so every later assertion proves a real transition, checks the cost-centre update doesn't clobber the earlier visibility flip, and confirms LAG-immutable attributes are unchanged. Uses `t.Parallel()` with `RequireSharedIntegrationClient` per the documented parallel pattern, and decommission/cleanup is delegated to the existing poll helper. Approving.

PR #453, opened 2026-06-22
test: add MCR/MVE lock/unlock lifecycle integration tests

Phil-Browne:
Adds integration coverage for resource-state mutations that previously had none: lock/unlock on MCR and MVE.

- **MCR lock/unlock** (`mcr_lifecycle_integration_test.go`): buy, lock, assert locked via an SDK read, unlock, assert cleared.
- **MVE lock/unlock** (`mve_lifecycle_integration_test.go`): same shape.

Notes:
- Both carry the `integration && provisioning` build tag, so the nightly read-only job never runs them. They run in the manual provisioning job, which already discovers the `mcr`/`mve` packages (no workflow change needed).
- Lock state is asserted through an SDK read because the CLI output doesn't surface the `Locked` field.
- Cleanup runs via `t.Cleanup` and unlocks before delete, so a mid-test failure never leaves a resource locked. No `time.Sleep`.
- Verified on staging: both lock tests pass (~50s each).

MVE restore was dropped from scope: staging rejects terminate-later (`CANCEL`) on an MVE (`409 Action [CANCEL] not allowed for product type: MVE`), so it can never reach the `CANCELLED` state that restore requires. The now-dead `mve restore` CLI command is tracked for removal separately.

ESD-1529

### DIFF
--- a/internal/commands/mcr/mcr_lifecycle_integration_test.go
+++ b/internal/commands/mcr/mcr_lifecycle_integration_test.go
@@ -1,0 +1,103 @@
+//go:build integration && provisioning
+
+package mcr
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/megaport/megaport-cli/internal/base/output"
+	"github.com/megaport/megaport-cli/internal/testutil"
+	megaport "github.com/megaport/megaportgo"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):
LGTM. Clean and symmetric across MCR and MVE: asserts the unlocked baseline before locking (so the post-lock assertion proves a real transition), reads lock state via SDK since the CLI doesn't surface it, and cleanup unlocks before delete (LIFO) since the API can refuse to delete a locked resource. MVE helper discovers the Aruba image dynamically rather than hard-coding a staging image ID, which avoids a known drift flake. No `time.Sleep`. These provision real MCR/MVE resources (~50s each) — inherent to provisioning lifecycle tests, not a concern. Approving.

PR #454, opened 2026-06-22
test: add MCR IPv6 prefix-filter-list integration lifecycle

Phil-Browne:
Extends MCR integration coverage with an IPv6 prefix-filter-list lifecycle test (ESD-1531). It buys an MCR, then creates/gets/updates/deletes an IPv6 prefix filter list and checks each step against authoritative SDK reads rather than scraped CLI output.

- `TestIntegration_MCRIPv6PrefixFilterLifecycle`: full create/get/update/delete on an IPv6 list, asserting address family, description, and per-entry action/ge/le via the SDK.
- Uses RFC 3849 documentation prefixes (`2001:db8::/32`) and canonical-prefix comparison so IPv6 normalization can't cause spurious mismatches. Cleanup is `t.Cleanup` in LIFO order (prefix list before MCR), no `time.Sleep`.
- Re-tags `mcr_integration_test.go` to `//go:build integration && provisioning` so the MCR lifecycle tests are kept out of the nightly read-only build at build level, not just by the `-run ReadOnly$` name filter.
- Documents the routing/BGP scope: there is no standalone CLI or SDK command to configure static routes or BGP on an MCR (that lives in VXC A-end partner config), so prefix filter lists are the only standalone MCR routing config covered here. Also corrects the build-tag rationale in `INTEGRATION_TESTING.md`.

Two notes for reviewers: the `ge`/`le` assertions compare exact values on the assumption staging echoes them back unchanged, so if staging ever normalizes them that is where a mismatch would surface (flagged in a code comment). The provisioning job is manual-dispatch only, so this test does not run on the PR; verification here was build/vet/lint/discovery across both tag combinations.


penzeliz-megaport (approved):
LGTM. Strong lifecycle test: SDK-backed assertions (not scraped CLI output), order-independent entry matching via `canonicalPrefix`, RFC 3849 documentation prefixes to avoid real-address collisions, delete verified by expecting a get error, and `t.Cleanup` registered LIFO before the UID assertion so a parse failure still tears down. One minor note: the delete-error assertion is loose (`assert.Error`, not specifically a 404/not-found), so a transient network error would also satisfy it — acceptable for a lifecycle smoke. The whole-file re-tag to `integration && provisioning` is the stated intent and matches the doc change. Approving.

PR #455, opened 2026-06-22
ESD-1532(fix): remove the non-functional mve restore command

Phil-Browne:
`mve restore` could never succeed. It calls the SDK's `RestoreProduct` (an un-cancel action), which needs a `CANCELLED` resource, but `DeleteMVE` only deletes immediately (`CANCEL_NOW` -> `DECOMMISSIONED`) and staging rejects the terminate-later `CANCEL` an MVE would need (`409 Action [CANCEL] not allowed for product type: MVE`, confirmed live in ESD-1529). This removes the dead command and everything that only served it.

- Drop the `restore` command builder and its registration in `mve.go`
- Remove `RestoreMVE`, `restoreMVEFunc`, and the WASM `restoreMVEWasmImpl` plus its init override
- Delete the two restore unit tests
- Regenerate docs (drop the `mve restore` rows) and delete the two orphaned doc files
- Add a `Removed` note to the changelog

MCR restore and the shared SDK `RestoreProduct` method are untouched.

ESD-1532


penzeliz-megaport (approved):

penzeliz-megaport (approved):
LGTM. Clean symmetric removal: command builder + registration, `RestoreMVE`/`restoreMVEFunc`/`restoreMVEWasmImpl` and the WASM init override, unit tests, generated docs + index rows, and a CHANGELOG `Removed` entry all go together, with no dangling references left behind. Given the command could never succeed (MVE never reaches `CANCELLED`, per ESD-1529), no working workflow is affected. Approving.

PR #457, opened 2026-06-23
ci: generate CHANGELOG.md from release notes, backfill to v0.13.0 (ESD-1533)

Phil-Browne:
Makes CHANGELOG.md generate from release notes instead of being hand-maintained, and catches it up through v0.13.0.

- Re-enable GoReleaser changelog generation: Added (feat), Changed (perf), Fixed (fix), plus an "Other Changes" catch-all so non-conventional `ESD-XXXX:` subjects aren't dropped. Docs/test/chore/ci/build/style/refactor, merges, and reverts are filtered out.
- New release-workflow job folds the generated notes into CHANGELOG.md via `scripts/update-changelog.sh` and opens a PR. It opens a PR rather than pushing to `main` because `main` is protected by org rulesets that require PRs.
- Backfill v0.6.0 through v0.13.0; every date and compare link verified against git. Pre-v0.6.0 entries are untouched.

One prerequisite for the automation: the org setting "Allow GitHub Actions to create and approve pull requests" must be enabled. If it is off, the job still pushes the `chore/changelog-<tag>` branch and logs a link to open the PR by hand, so a release is never blocked.

v0.5.6 through v0.5.9 are folded into the v0.6.0 entry (its compare link spans them); the ticket scoped the catch-up to v0.6.0 onward.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -71,3 +71,54 @@ jobs:
           subject-path: |
             dist/megaport-cli_*.zip
             dist/megaport-cli_*_SHA256SUMS
+
+  # Fold the notes GoReleaser generated for the draft release into CHANGELOG.md
+  # and open a PR against the default branch (which is protected by org rulesets,
+  # so a direct push is not possible). Runs after goreleaser so a failure here
+  # never blocks the build/sign/publish steps above.
+  update-changelog:
+    name: Update CHANGELOG.md
+    runs-on: ubuntu-latest
+    needs: goreleaser
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5.0.1
+        with:
+          ref: ${{ github.event.repository.default_branch }}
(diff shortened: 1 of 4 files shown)


penzeliz-megaport (approved):
LGTM. The `update-changelog` job is scoped correctly: `contents: write` + `pull-requests: write` live on this one job (not the whole workflow), it runs only as part of the release flow (`needs: goreleaser`, tag-driven, not reachable from PR triggers), uses the default `GITHUB_TOKEN` rather than a PAT, and untrusted inputs (`TAG`/`PREV_VERSION`) are passed via `env`. `actions/checkout` is SHA-pinned.

Two operational notes (not defects, none blocking):
1. `git push --force` targets the bot-owned `chore/changelog-<tag>` branch only — low blast radius, but a deliberate force-push worth being aware of.
2. The auto-PR step depends on the org setting "Allow GitHub Actions to create and approve pull requests"; the diff confirms it degrades gracefully (logs a manual-PR link) if that's off.

I didn't execute the awk/sed transform, so CHANGELOG output shape for every release-note variant isn't certified here — worth a glance on the first real release. Approving.

PR #463, opened 2026-06-29
fix(vxc): accept valid single-field flag updates (ESD-1550)

Phil-Browne:
`megaport vxc update <uid>` rejected valid single-flag updates (for example `--shutdown`, `--cost-centre`, `--term`, `--is-approved`, or the vNIC index flags) with "at least one field must be updated". The non-interactive gate checked a stale flag list: it omitted most of the flags `WithVXCUpdateFlags` registers and included three that are not update flags at all (`a-end-location`, `b-end-location`, `locked`).

The fix replaces the hardcoded list with a single source of truth, `cmdbuilder.VXCUpdateFlagNames`, that the gate iterates, so it can no longer disagree with the flags the command registers.

Changes:
- New `cmdbuilder.VXCUpdateFlagNames` lists exactly the flags `WithVXCUpdateFlags` registers; the gate iterates it.
- Drop the three dead entries from the gate.
- Table test: each update flag set on its own passes the gate, no flags fails, and a non-update flag does not pass.
- Drift-guard test asserts `VXCUpdateFlagNames` matches what `WithVXCUpdateFlags` actually registers, so the two cannot drift again.

Fixes ESD-1550.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ workflow (scripts/update-changelog.sh). Don't hand-edit them or add entries unde
 - vRouter interface validation now accepts an untagged VLAN (-1), matching the Azure peer and shared VLAN checks. Previously a `-1` VLAN was rejected client-side even though the API accepts it, which also blocked the interactive "press Enter for no VLAN" path
 - The interactive BGP connection password is now read without echo, matching the IPsec pre-shared key. Previously it was typed in cleartext on screen
 - Interactive `vxc update` now applies the A-End vRouter partner config to the A-End. Previously it was written to the B-End field, so the A-End config was lost and the B-End was overwritten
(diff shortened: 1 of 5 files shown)


penzeliz-megaport (approved):
LGTM. Verified `VXCUpdateFlagNames` matches exactly the 16 flags `WithVXCUpdateFlags` registers (4 common + 6 endpoint + 2 partner-config + shutdown/is-approved/a-vnic-index/b-vnic-index), and the builder does not pull in `WithResourceTagFlags`, so no stray flags leak into the gate. The single-source-of-truth + drift-guard test is the right approach.

One thing to confirm (non-blocking): `emptyMeansUnsetUpdateFlags` only special-cases the partner-config flags, so `--cost-centre ""` still passes the gate and clears the cost centre. That matches the builder's intent, just flagging it's deliberate.

PR #464, opened 2026-06-29
ESD-1549: fix vxc buy dropping --resource-tags on the flags path

Phil-Browne:
`vxc buy --resource-tags '{"env":"prod"}'` built the order from flags and silently dropped the tags: the flags builder never read the flag, while the `--json` path applied it. This wires the flag in so both paths set tags identically.

- `buildVXCRequestFromFlags` now reads `--resource-tags` and assigns it to the buy request, matching the JSON path.
- Parsing goes through a new shared `utils.ParseResourceTagsFlag` helper (json parse + empty-key rejection), so the two paths converge on one parse and one validation.
- Malformed `--resource-tags` JSON now returns a usage error (exit 2) before any order call, instead of being misclassified.
- Tests: flags-path round-trip, nil-when-unset, malformed-JSON usage error, and empty-key rejection, plus direct helper coverage.

Follow-up (out of scope, pre-existing): `--resource-tags-file` is registered on `vxc buy` (and ports/mcr) but unwired on every input path. Worth a separate ticket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ workflow (scripts/update-changelog.sh). Don't hand-edit them or add entries unde
 - Bumped the megaportgo SDK to v1.14.1, which models `ipSecTunnelOptions` as a single object per `ipSecTunnel` interface (one tunnel each) instead of an array
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


penzeliz-megaport (approved):
LGTM. Wiring `--resource-tags` into `buildVXCRequestFromFlags` and converging both paths on `TagMapFromObject` is clean. Confirmed the JSON-path refactor is behaviour-preserving: it keeps the same non-string and empty-key rejection and the identical error string, so no regression there. Malformed JSON now returns a usage error before the order is placed, which is correct. Test coverage is solid.

PR #465, opened 2026-06-29
ESD-1556: wire resource-tag flags into the buy flags builders

Phil-Browne:
Stacked on #464 (ESD-1549). Review/merge that first; this PR targets its branch so the diff shows only the ESD-1556 changes.

`WithResourceTagFlags` registers `--resource-tags` and `--resource-tags-file` on every buy command, but the flags builders only honored some of them: `mve buy` dropped both, and `vxc`/`ports`/`mcr` ignored `--resource-tags-file`. The flags were registered and documented, so the user got no signal they did nothing.

This wires all four buy flags builders through one shared helper.

- New `utils.ParseResourceTagsFlagOrFile`: reads the `--resource-tags` JSON string or, when empty, the `--resource-tags-file` path (string wins, mirroring `utils.ReadJSONInput`), validating through `TagMapFromObject` so the flags and JSON paths accept and reject identically.
- `vxc`, `ports` (both port and LAG builders), `mcr`, and `mve` buy now use it. MVE was net-new (it read neither flag).
- Malformed JSON, non-string values, and empty keys return a usage error (exit 2) before any order is placed.
- Unit tests: file-path round-trip per resource type, MVE flags-path round-trip, string-over-file precedence, and malformed-input usage errors.

Out of scope: JSON/interactive paths, `nat-gateway` (the reference), and update commands are untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ workflow (scripts/update-changelog.sh). Don't hand-edit them or add entries unde
 
 ### Fixed
 - `vxc buy --resource-tags` now applies tags when buying via flags, matching the `--json` path. Previously the flag was registered but silently ignored. Both paths now share one parse and validation, so malformed JSON, non-string values, and empty keys return a usage error before the order is placed
+- `mve buy --resource-tags` / `--resource-tags-file` and the `--resource-tags-file` flag on `vxc`, `ports`, and `mcr` buy now apply tags on the flags path, matching `nat-gateway`. Previously these registered flags were silently ignored; malformed JSON, non-string values, and empty keys now return a usage error before the order is placed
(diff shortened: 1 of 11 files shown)


penzeliz-megaport (approved):
LGTM. Now that #464 is merged, the retarget to `main` is clean: the diff shows only the ESD-1556 changes with no duplication of #464's work, and the CHANGELOG entry slots in below the vxc line with no conflict.

The shared `ParseResourceTagsFlagOrFile` reuses the hardened `readTagsFile` (path-traversal rejection, regular-file check, 1 MiB cap, TOCTOU-safe), with string-over-file precedence matching `ReadJSONInput`. All four buy builders (vxc, ports + LAG, mcr, mve) now converge on it; MVE was genuinely net-new. Test coverage is thorough — per-resource file round-trips, precedence, and malformed-input usage errors.

All 10 check runs are green on this head (Unit Tests, lint, Build Guard, govulncheck, hermetic E2E on both OSes, codecov). Note the checks ran before the base retarget; head is unchanged so they remain valid, but a re-run against current `main` would make it fully airtight if branch protection requires it.

PR #466, opened 2026-06-30
docs: prepare documentation for beta release

Phil-Browne:
One more docs pass ahead of the beta release.

- Flip the README status note from alpha to beta.
- Fix the portal-to-CLI migration guide: the bulk port-buy script used `--force`, but `ports buy` skips its prompt with `--yes`. As written it would error with "unknown flag".
- Document the `--fields` and `--query` output flags in the README (both are used in the guides but were undocumented).
- Bump the stale Docker pin example from v0.4.7 to v0.13.0.
- Regenerate command docs so they match the code: `apply` now explains the per-resource `--timeout`, and `nat-gateway create` lists the `--asn` flag.
- Remove two dev planning/spec files that leaked into `docs/superpowers/`.

All command and flag references in the guides were checked against the built binary.


mp1-tomclancy (approved):

projectsuperba (approved):
`docs/superpowers/` will probably remain on any dev machines without a .gitignore but not really a show stopper

PR #467, opened 2026-06-30
test(integration): add location and MVE capacity fallback (ESD-1559)

Phil-Browne:
Staging keeps running out of capacity at the location our integration tests pinned (67), so the MVE tests (and others) intermittently fail with `400 No available capacity`. This makes the provisioning tests pick a usable location at run time instead.

- Shared finders in `testutil` prefer the historical location, else the first active location advertising the needed speed, else skip.
- MVE host capacity isn't visible in advertised speeds, so it probes `ValidateMVEOrder` live and falls back across product sizes before skipping.
- The ports, MCR, VXC, MVE, service key, and apply provisioning tests now use the finders.
- Verified on staging: ports 6/6, VXC 6/6, MCR 6/6, MVE 3/3, service keys 3/3, apply 3/3 green. The MVE fallback fired live: the preferred location was out of capacity, so the test bought at the fallback location.

Runner passes fully in CI https://github.com/megaport/megaport-cli/actions/runs/28445324839

### DIFF
--- a/internal/commands/mve/mve_integration_test.go
+++ b/internal/commands/mve/mve_integration_test.go
@@ -16,13 +16,16 @@ import (
 	"github.com/megaport/megaport-cli/internal/base/output"
 	"github.com/megaport/megaport-cli/internal/testutil"
 	"github.com/megaport/megaport-cli/internal/validation"
+	megaport "github.com/megaport/megaportgo"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-// stagingMVELocationID is a known MVE-capable staging location. Discovering one
-// via ListLocations would be overengineering for this lifecycle test.
+// stagingMVELocationID is tried first when probing for MVE capacity. It is not
+// authoritative: when it is out of capacity, findMVECapacity sweeps every other
+// active MVE-capable location before giving up (mirroring the terraform
+// provider's acceptance tests).
 const stagingMVELocationID = 65
 
 func generateUniqueID() string {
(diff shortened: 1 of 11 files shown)


internal/commands/mve/mve_integration_test.go
```diff
@@ -144,20 +147,154 @@ func discoverArubaImage(t *testing.T) discoveredImage {
...
+				ProductSize: size,
+				AccountName: "test",
+				AccountKey:  "test",
+				SystemTag:   "test",
+			},
+			Vnics: []megaport.MVENetworkInterface{
+				{Description: "MVE VNIC 1", VLAN: 55},
+				{Description: "MVE VNIC 2", VLAN: 56},
+			},
+		}) == nil
```
mp1-tomclancy: Could this check also use `isMVECapacityError`, rather than assuming `err != nil` is equivalent to no capacity?

internal/testutil/integration.go
```diff
@@ -159,3 +159,91 @@ func SharedIntegrationClient(t *testing.T) *megaport.Client {
...
+
+// FindPortTestLocation returns a staging location that advertises port capacity
+// at speedMbps, preferring preferredID. See findOrderableLocation for the
+// fallback and skip behavior.
+func FindPortTestLocation(t *testing.T, client *megaport.Client, speedMbps, preferredID int) int {
+	t.Helper()
+	return findOrderableLocation(t, client, preferredID, fmt.Sprintf("%d Mbps port capacity", speedMbps), func(loc *megaport.LocationV3) bool {
+		return locationHasPortSpeed(loc, speedMbps)
+	})
+}
```
mp1-tomclancy, line 230: I'm guessing that `FindPortTestLocation` and `FindMCRTestLocation` skip the full validation workflow that  `findMVECapacity` uses for performance reasons?

mp1-tomclancy (dismissed):
Looks good. Added some non-blocking comments/questions.

internal/testutil/integration.go
Phil-Browne: Yup, port/MCR capacity is exposed directly in the location's advertised speeds (locationHasPortSpeed/locationHasMCRSpeed), so a static filter over ListLocationsV3 is enough for now.

internal/commands/mve/mve_integration_test.go
Phil-Browne: Addressed in https://github.com/megaport/megaport-cli/pull/467/commits/2d80e87136099ccf57fa5cca936a6314ed3691eb

mp1-tomclancy (approved):

PR #474, opened 2026-07-05
chore(ESD-1574): hide docs subcommand and drop generated twin pages

Phil-Browne:
The CommandBuilder auto-adds a `docs` subcommand to every command. Because it wasn't hidden, it cluttered every command's help listing and `generate-docs` emitted a redundant `megaport-cli_<cmd>_docs.md` twin page for each one: 139 identical boilerplate pages that were embedded into the binary but never read (the `docs` subcommand renders its parent's page, not a twin).

This hides the auto-added `docs` subcommand, which fixes both problems at the source: `generate-docs` already skips hidden commands, so the twins stop generating, and `docs` no longer shows in help listings. `megaport-cli <cmd> docs` still works if typed.

- Set `Hidden: true` on the auto-added docs subcommand.
- Drop the 139 stale `_docs.md` twin pages.
- Regenerate the command pages (each loses its `docs` subcommand bullet) and trim the 139 twin rows from `index.md`.
- Pin the behavior with a `Hidden` assertion in `TestBuild`.

The `index.md` header (date/version) is left as-is to keep this scoped to twin removal, not a docs version bump. Man-page generation gets the same twin fix for free.


penzeliz-megaport (approved):
CI green. Hidden:true on the auto-added docs subcommand is the right fix at source, and the twin-page removal follows. One coordination note: this overlaps #476 on docs/index.md (and the regenerated mve command pages). Suggest merging this first, then rebasing #475/#476 and re-running generate-docs so the docs version stamp is deliberate rather than a merge-order artefact. LGTM.

PR #483, opened 2026-07-06
ESD-1580-fix(wasm): keep status messages when table output is present

Phil-Browne:
Browser CLI users got a blank table with no explanation on empty results: `GetCapturedOutput` returned a single output source by priority (JSON > CSV > XML > table > direct), so status messages (`PrintInfo`/`PrintWarning`/`PrintSuccess`/`PrintPlain`) written to the direct buffer were dropped whenever a table was present. The "No X found" warning was silently lost behind a header-only table.

In table mode we now prepend the direct status buffer to the table output, so status lines appear above the table, matching native's separate stderr/stdout streams. JSON/CSV/XML stay data-only.

Changes:
- `GetCapturedOutput`: table branch returns `direct + tableOutput` (status first, then table).
- Empty slices still render a header-only table (native parity); the warning now surfaces above it.
- WASM-tagged tests for the empty-result and status-plus-table cases, plus a table-driven test that JSON/CSV/XML stay clean.
- Tagged the native-only output test files `//go:build !wasm` (they use `os.Pipe`-based capture helpers, unimplemented in WASM) and moved shared fixture structs to an untagged helper, so the output package's WASM test build compiles and runs.

ESD-1580


penzeliz-megaport (dismissed):
Approving. The fix is correct in isolation, and the #489 overlap is a one-line rebase note — keep the `direct + table` concat in the no-streaming fallback when whichever PR lands second rebases — not a merge-order blocker. CI-coverage note left in the comment above.

penzeliz-megaport (dismissed):
Re-approving (dismissed by the 07/07 CI commit, which is the step I asked for — thanks).

The output-priority fix and its `TestGetCapturedOutput_*` regression tests are sound. Two rebase-time items, neither blocking:

1. Branch predates #484, so it lacks main's `make wasm-vet` step and reverts its comment block — restore both on rebase.
2. `TestGetCapturedOutput_Priority` lives in `internal/wasm/wasm_test.go`, which #489 re-tags `js && wasm && browser`. If #489 lands first, this regression stops running under the default headless CI. I've left the detail on #489 (comment: split the node-safe tests out of that mixed file). Worth coordinating so the priority test survives whichever order these merge.

penzeliz-megaport (dismissed):
Re-approving. The merge of main pulled in #484's `wasm-vet` step and #487's wasmhttp step, so the behind-#484 concern is resolved here. `wasm_test.go` stays `//go:build js && wasm` with `TestGetCapturedOutput_Priority` runnable, and the new commit only removes the unused `calculateColumnWidths`.

Now that #489 has split the browser-only tests into `wasm_browser_test.go`, the coverage trap I flagged is gone: once these land, this PR's `-run 'TestGetCapturedOutput' ./internal/wasm ...` step is redundant with #489's whole-package `./internal/wasm` run and could be dropped — not required, just tidy-up.

penzeliz-megaport (approved):
Re-approving. The dismissing commit deletes only the hand-edited `CHANGELOG.md` line for this ticket — the file header explicitly forbids hand-edits (entries are regenerated by `scripts/update-changelog.sh`), so this is correct hygiene, not a loss: the entry will be auto-generated at release. The code I previously reviewed is unchanged and CI is fully green.

PR #495, opened 2026-07-06
ESD-1601-refactor(wasm): remove the config manager from the browser build

Phil-Browne:
Removes the dead WASM-specific `ConfigManager` (a browser-localStorage credential store nothing in the browser build ever reached) and tags its remaining native-only callers so they drop out of the `js,wasm` compile graph. This also fixes a real latent build break: `GOOS=js GOARCH=wasm go build ./...` failed because `auth_actions.go` called `GetProfile`, a method that only ever existed on the native `ConfigManager`.

- Deleted `manager_wasm.go` and four now-orphaned WASM shims with zero remaining callers: `config_actions_wasm.go`, `config_module_wasm.go`, `config_wasm.go`, `storage_wasm.go`
- Tagged `config_actions.go`, `config_utils.go`, `config.go`, `config_module.go`, and the whole `auth` package `//go:build !js && !wasm` (matching `manager.go`/`storage.go`, which were already native-only)
- No native or WASM user-facing behavior changes: `config` and `auth` were already unreachable in the browser build before this change

Verified clean: native build/test/lint, `GOOS=js GOARCH=wasm go build ./...`, and `go vet -tags js,wasm ./...` (the one remaining `GetProfile`-related vet failure is gone).

**Known pre-existing gap, out of scope here:** `internal/commands/config/login_test.go` mixes native and WASM tests in one untagged file and already failed `go vet -tags js,wasm ./...` on `main` before this change (`undefined: appendLogOpts`); after this change it fails earlier in the same file (`undefined: NewConfigManager`) for the same underlying reason. Not exercised by CI (`wasm.yml` only runs `go test -tags js,wasm ./cmd/megaport`). Belongs to the wasm build/vet/lint CI hardening epic (ESD-1578).


penzeliz-megaport (dismissed):
Approving. Removing the browser config manager is clean: verified no dangling Go references in the wasm build (the localStorage hit was a JS-name match, not a Go call), and the native build is unaffected. Net security improvement — drops a plaintext access-key/secret-key localStorage store. Minor in-scope tidy: `login_test.go` still references deleted symbols under `go vet -tags js,wasm` (one-line `//go:build !js && !wasm` fix).

penzeliz-megaport (dismissed):
Re-approving (dismissed by the 07/07 commit).

The delta just adds `//go:build !js && !wasm` to `login_test.go`, which is correct: this PR removes the config manager (and login) from the browser build, so its test must be excluded from the wasm target rather than left to fail compilation there. Consistent with the rest of the PR's build-tag alignment.

penzeliz-megaport (dismissed):
Re-approving (dismissed by the 23:17 merge of main). Delta is a merge-of-main catching the branch up after #482/#484/#487 landed; the config-manager removal and the `login_test.go` build-tag exclusion I reviewed are unchanged.

penzeliz-megaport (approved):
Re-approving — and the wasm build fix is correct. The follow-up commit moves the `ConfigManager` struct out of the shared `config_shared.go` into native-only `manager.go` (`//go:build !js && !wasm`), where its `NewConfigManager` constructor already lives. Verified: all 12 `ConfigManager` references sit in native-only files (`config_actions.go`, `manager.go`); the shared file and both wasm files reference it zero times, so the browser build no longer carries the config manager — exactly the intent of ESD-1601. CI fully green: Build Guard, lint (js/wasm), and govulncheck all pass.

PR #507, opened 2026-07-08
ESD-1615: add matrix-aware NAT gateway speed/session validation

Phil-Browne:
NAT gateway `create` only checked that speed and session count were positive, so an unsupported speed/session combination reached the API instead of failing pre-flight. This bumps `megaportgo` to v1.15.0 and uses its `NATGatewaySpeedSessionSupported` helper to check the pair against the live availability matrix before ordering, listing the valid speeds and session counts if it's rejected.

- Fetch the matrix via `ListNATGatewaySessions` and validate the requested speed/session pair against it in `nat-gateway create`
- Fail open on a matrix fetch failure or empty matrix (the API still enforces the rule at order time), matching the Terraform provider's handling
- Add stub mock methods for the new MCR ping/traceroute interface methods that came in with the same SDK bump, so `mcr` keeps compiling (the commands themselves are ESD-1614)

### DIFF
--- a/internal/commands/nat_gateway/nat_gateway_actions.go
+++ b/internal/commands/nat_gateway/nat_gateway_actions.go
@@ -128,6 +128,9 @@ func validateNATGatewaySpeedSessionMatrix(ctx context.Context, client *megaport.
 	if !result.SpeedSupported {
 		return validation.NewValidationError("speed", req.Speed, fmt.Sprintf("not supported; supported speeds (Mbps): %v", result.SupportedSpeeds))
 	}
+	if req.Config.SessionCount == 0 {
+		return nil
+	}
 	return validation.NewValidationError("session count", req.Config.SessionCount,
 		fmt.Sprintf("not supported for %d Mbps; valid session counts: %v", req.Speed, result.SessionsAtSpeed))
 }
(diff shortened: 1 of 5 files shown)


internal/commands/nat_gateway/nat_gateway_actions.go
```diff
@@ -101,6 +106,32 @@ func CreateNATGateway(cmd *cobra.Command, args []string, noColor bool) error {
...
+		return nil
+	}
+	if len(matrix) == 0 {
+		return nil
+	}
+
+	result := megaport.NATGatewaySpeedSessionSupported(matrix, req.Speed, req.Config.SessionCount)
+	if result.Supported {
+		return nil
+	}
```
mp1-tomclancy, line 127: When`req.Config.SessionCount = 0` (i.e. because `--session-count` is an optional int) I'm guessing that validation would fail. Might need to bypass and let it fall through to the API validation in this case?

mp1-tomclancy (changes requested):

internal/commands/nat_gateway/nat_gateway_actions.go
Phil-Browne: Fixed in https://github.com/megaport/megaport-cli/pull/507/commits/ba7d40bbd6ae6d990e1e638ee5285bc835500ef1 thanks!

mp1-tomclancy (approved):

PR #512, opened 2026-07-10
ESD-1626: Fix interactive VXC buy dropping the B-End VLAN

Phil-Browne:
The interactive VXC buy prompt collected a B-End VLAN, then wrote it to the request's zero-value `BEndConfiguration`. A few lines later that struct is overwritten by the assembled `bEndConfig`, so the entered VLAN was silently dropped and the API assigned one instead. The flag and JSON buy paths were unaffected.

- Assign the entered B-End VLAN to `bEndConfig` (the struct actually submitted), matching the A-End path.
- Add regression tests: an entered B-End VLAN now survives onto the request, and skipping the prompt leaves it unset.


penzeliz-megaport (approved):

PR #513, opened 2026-07-10
ESD-1624: Stop trusting the page-writable apiURL global in WASM login

Phil-Browne:
The WASM login path read the outbound API base URL from `window.megaportToken.apiURL`, a page-writable JS global. Any other script on the page (XSS, a compromised widget, a browser extension) could overwrite it after login and redirect the bearer token to an attacker host, defeating the hostname validation `setAuthToken` already does.

Both login call sites now read the URL from the `MEGAPORT_API_URL` env var that `setAuthToken` sets after validating the portal hostname. The `apiURL` field on the JS global stays for UI display only, since nothing security-sensitive reads it back.

- `loginFunc` (token path) and `newUnauthenticatedClientFunc` now source the base URL from `os.Getenv("MEGAPORT_API_URL")`.
- New `js,wasm` tests assert both paths route to the validated host even when the JS global is tampered to an attacker URL.
- Added a `wasm.yml` step so the config package's wasm tests actually run under node in CI (previously compiled but never executed).


internal/commands/config/login_wasm.go
```diff
@@ -93,7 +93,11 @@ var loginFunc = func(ctx context.Context) (*megaport.Client, error) {
 	if !megaportTokenGlobal.IsUndefined() && !megaportTokenGlobal.IsNull() {
 		token := os.Getenv("MEGAPORT_ACCESS_TOKEN")
 		tokenEnv := megaportTokenGlobal.Get("environment").String()
```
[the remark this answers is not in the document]
Phil-Browne: Good catch, agreed. `tokenEnv` now reads from `MEGAPORT_ENVIRONMENT` (the bucket `setAuthToken` stores alongside `MEGAPORT_API_URL`) and only falls back to the JS global if that env var is empty. This keeps the page-writable global out of the fallback host selection and debug logging, matching AC #3. Added `TestLoginFallbackEnvIgnoresTamperedGlobal` to cover the fallback path with a tampered global. Fixed in cd3737f5.

penzeliz-megaport (dismissed):

internal/commands/config/login_wasm.go
projectsuperba: Perhaps out of scope but related: should this environment value also prefer the env flag over the global?

projectsuperba (dismissed):

internal/commands/config/login_wasm.go
Phil-Browne: Agreed, done in 55e59860. The unauthenticated client fallback now checks MEGAPORT_ENVIRONMENT first (set by both setAuthToken and setAuthCredentials) and only consults the page-writable megaportCredentials global when the env var is empty, matching the tokenEnv ordering in loginFunc. Added tests covering both the tampered-global case and the global-only fallback.

projectsuperba (approved):

PR #519, opened 2026-07-10
ESD-1629: emit WASM artifact integrity hashes from the publish workflow

Phil-Browne:
Extends the WASM publish workflow to emit integrity reference values and attest the artifact, so the Portal can verify the `megaport.wasm` it loads from media.megaport.com is the one we published (SEC-8095).

- New "Compute integrity hashes" step: SHA-256 (hex) of the uncompressed `megaport.wasm` and the `sha384-<base64>` SRI of `wasm_exec.js`.
- The step summary now shows both values under the version prefix, ready to copy into the Portal WASM-loader config per release.
- Adds an `actions/attest-build-provenance` step (SHA-pinned, same version as `release.yaml`) for `megaport.wasm`, plus `attestations: write` on the job.

The hash is taken over the pre-brotli bytes, i.e. what the browser sees after the `Content-Encoding: br` decode, not the `.br` object. Verifying the deployed artifact with `gh attestation verify` must therefore be done against the br-decoded bytes, not the raw CDN object.

No change to the S3 publish, cache-control, or any Portal/web-monorepo code.

Relates to SEC-8095. Blocks the Portal WASM-loader integrity check.

### DIFF
--- a/.github/workflows/wasm-publish.yaml
+++ b/.github/workflows/wasm-publish.yaml
@@ -42,6 +42,14 @@ name: Publish WASM to S3
 #   write access. The shared deploy role trusts the whole repo, so to add a required-
 #   reviewers approval gate, move this job into a `production` GitHub environment (needs a
 #   repo admin to create it) and re-add `environment: production` to the job.
+#
+# INTEGRITY (SEC-8095):
+#   The build-provenance attestation proves the published megaport.wasm came from this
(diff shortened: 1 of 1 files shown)


penzeliz-megaport (dismissed):
Reviewed for correctness; findings noted in comments are non-blocking. LGTM.

ianb-mp (approved):

PR #520, opened 2026-07-10
ESD-1633-fix(wasm): stop SplitArgs deleting a literal megaport argument

Phil-Browne:
Fixes a WASM-only bug where `SplitArgs` dropped any argument equal to `megaport`, `megaport-cli`, or `./megaport-cli`, not just a leading program name. That meant a command like `ports update <uid> --name megaport` silently lost the flag value.

- Only strip a program-name token when it's the first token, leaving later flag values and positionals intact.
- Added test cases covering a `megaport`-valued flag, a positional, and a leading token combined with a later literal value.

Ran `go build`, `go test ./...`, the WASM build, WASM package tests, and lint locally; all pass.

### DIFF
--- a/internal/wasm/wasm.go
+++ b/internal/wasm/wasm.go
@@ -736,14 +736,12 @@ func SplitArgs(cmd string) []string {
 		args = append(args, currentArg.String())
 	}
 
-	// Remove program name if user included it.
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):

projectsuperba (approved):

PR #523, opened 2026-07-10
ESD-1636-fix(wasm): recover around JS callback invocations at the boundary

Phil-Browne:
Two JS callback invocations at the WASM boundary weren't guarded against a throwing host callback: the prompt callback and the async command timeout callback. A throwing callback here previously either leaked the pending prompt entry or could crash the whole WASM runtime.

- Added an `InvokeCallback` helper (`internal/wasm/callback.go`) that recovers around `js.Value.Invoke`, mirroring the existing guard in `pushOutputChunk`
- Wired it into the prompt callback in `internal/wasm/prompts.go`, cleaning up the pending prompt entry if the callback throws
- Wired it into the async command timeout callback in `main_wasm.go`, extracted into `invokeAsyncTimeoutCallback` for testability
- Added tests covering both recover paths

The recovered panic value is deliberately never formatted into a log or error message. `js.Error.Error()` reads a JS `message` property back from the runtime, and property access has no recover boundary in `wasm_exec.js` (unlike Invoke), so formatting it could crash the runtime in a way Go can't catch. Fixed, friendly strings are logged instead.

Out of scope per the ticket: prompt/timeout durations, and the goroutine's own existing recover in `executeMegaportCommandAsync`, which isn't touched here.

### DIFF
--- a/internal/wasm/callback.go
+++ b/internal/wasm/callback.go
@@ -6,13 +6,21 @@ import "syscall/js"
 
 // InvokeCallback invokes cb with args, recovering from a panic so a throwing
 // host JS callback cannot escape into the caller. This mirrors the recover
-// pushOutputChunk (see output_handler.go) already applies around its own JS
-// Invoke. It returns the recovered panic value, or nil if the invocation
-// completed normally.
+// that pushOutputChunk (see output_handler.go) already applies around its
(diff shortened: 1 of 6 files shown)


penzeliz-megaport (approved):

PR #524, opened 2026-07-10
ESD-1635: Validate the environment in WASM setAuthCredentials

Phil-Browne:
`setAuthCredentials` wrote its environment argument straight into `MEGAPORT_ENVIRONMENT` with no validation. Downstream, `login_wasm.go`'s credential login path defaulted any unrecognized environment to the production API endpoint, so a typo, wrong case, or stray whitespace would silently authenticate real credentials against production instead of the intended environment.

- `setAuthCredentials` now normalizes (lowercase, trim) and validates the environment the same way `setAuthToken` does, rejecting anything outside production/staging/development instead of storing it.
- `login_wasm.go`'s credential path now fails closed on an unrecognized environment instead of defaulting to production.
- `login_wasm.go`'s bearer-token login path had the same fail-open default; it now fails closed too instead of silently routing an unrecognized token environment to production.
- The credential path also let the page-writable `megaportCredentials` JS global override an already-validated `MEGAPORT_ENVIRONMENT`, so a script sharing the page origin could redirect valid credentials to a different environment. It now only falls back to the global when the env var is empty, matching the token path's existing precedence.
- Added tests covering rejection, normalization, the fail-closed switch default on both login paths, and the global-override precedence fix.

Ref: ESD-1635


penzeliz-megaport (dismissed):

PR #525, opened 2026-07-10
ESD-1640: Read VXC shutdown status from the Shutdown field

Phil-Browne:
The VXC interactive update prompt showed the current shutdown status from `AdminLocked` instead of `Shutdown`. Those are separate fields on the SDK VXC type, so a VXC that was shut down but not admin-locked displayed as enabled, which could lead someone to toggle shutdown the wrong way.

- Read `vxc.Shutdown` instead of `vxc.AdminLocked` for the displayed shutdown status
- Added a test covering the shutdown-but-not-locked case and the inverse

The write path already used `req.Shutdown` correctly, so this is a display-only fix.


penzeliz-megaport (dismissed):

ianb-mp (approved):

PR #534, opened 2026-07-10
ESD-1650: Polish WASM output composition and HTTP transport fidelity

Phil-Browne:
Four fidelity gaps in the WASM build. The command tree stays alive across invocations in a browser session, so repeated `--help` calls kept re-coloring and re-suffixing the same help text. Structured output (JSON/CSV/XML/table) was overwritten on every print call, so a command emitting more than one document only kept the last. The WASM HTTP transport dropped every header value but the first and sent bodies as Go strings, which would mangle a binary body. And table output sliced the already-sanitized buffer by a raw byte offset, which stops lining up once sanitization strips enough control-sequence bytes, causing garbled output or a panic.

- Cache each command's pristine help text so repeated `--help` always renders from a clean base.
- Accumulate structured output buffers across calls within one command, clearing only between invocations.
- Join multi-value headers instead of dropping all but the first, and send request bodies as raw bytes via a Uint8Array.
- Roll back a structured buffer to its pre-call length if an encode fails partway through, so a failed call can't leak partial output into a later successful one.
- Slice the raw table buffer first and sanitize each piece separately, instead of slicing an already-sanitized buffer by a raw byte offset.

Tests added for the help-text regression, buffer accumulation (including a first-ever XML output test), the transport header/body fixes, and the table-output offset misalignment.


penzeliz-megaport (dismissed):

PR #538, opened 2026-07-11
fix(validation): correct int-width bound in GetIntFromInterface

Phil-Browne:
Extracted from #535 (ESD-1648) as a standalone fix, orthogonal to the prompt/validator-parity work in that PR.

`GetIntFromInterface` compared a JSON-derived `float64` against `math.MaxInt` directly. `float64(math.MaxInt)` rounds up to 2^63 only when `int` is 64 bits, so on platforms where `MaxInt` fits exactly in a float64 (e.g. 32-bit `int`, js/wasm) the check wrongly rejected the true max int.

- Use `>= float64(math.MaxInt)+1` as the exclusive upper bound, correct on either int width.
- Derive the boundary tests from `math.MaxInt` instead of a hardcoded 2^63 so they stay in range on any platform.

### DIFF
--- a/internal/validation/conversion.go
+++ b/internal/validation/conversion.go
@@ -14,9 +14,15 @@ func GetIntFromInterface(value interface{}) (int, bool) {
 		return v, true
 	case float64:
 		// Reject fractional or out-of-range values rather than silently truncating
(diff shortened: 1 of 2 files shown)


penzeliz-megaport (approved):

PR #539, opened 2026-07-11
fix(cli): handle empty-value tag edits in interactive tag prompt

Phil-Browne:
Extracted from #535 (ESD-1648) as a standalone fix, orthogonal to the validator-parity work in that PR.

The interactive tag editor keyed removal on `value == "" && tags[key] != ""`, which had two bugs:

- It refused to remove a tag whose current value was already empty.
- Entering an empty value for a key that did not exist silently dropped the key with no feedback.

Now key existence is checked explicitly: an empty value removes the tag if present, otherwise reports "nothing to remove". Fixed in both the native (`prompts.go`) and WASM (`prompts_wasm.go`) paths, which had drifted out of sync.

### DIFF
--- a/internal/utils/prompts.go
+++ b/internal/utils/prompts.go
@@ -252,11 +252,15 @@ var updateResourceTagsPromptFn = func(existingTags map[string]string, noColor bo
 				return nil, err
 			}
 
-			// Empty value means remove the tag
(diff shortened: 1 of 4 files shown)


penzeliz-megaport (approved):

